PR #26, opened 2022-09-21
docs fix on contributing link and menu bar shortcut

louis-csm:


clement-bonnet:
Thank you @louis-csm for your time and the solution you are suggesting.

Unfortunately, this solution does not completely resolve the issue as it does not work on GitHub. There is a discrepancy between how GitHub and mkdocs render emojis with markdown.

Closing for now, until a solution is found.

Issue #27, opened 2022-09-22
feat: expose state and key as attributes in DMEnvWrapper


clement-bonnet:
### Is your feature request related to a problem? Please describe
When working with a `dm_env.Environment` version of a Jumanji environment using the `JumanjiToDMEnvWrapper`, one may need to get/set the state and key of the environment e.g. to allow planning and "restart" the environment to its previous state.

### Describe the solution you'd like
Right now, this is possible by calling `wrapped_env._state` and `wrapped_env._state` which should not be allowed since key and state are private attributes.
A solution would be to properly expose them as properties and to implement setters for these properties (in the [common style](https://stackoverflow.com/questions/2627002/whats-the-pythonic-way-to-use-getters-and-setters#:~:text=%40property%0A%20%20%20%20def%20x(self)%3A%0A%20%20%20%20%20%20%20%20%22%22%22I%27m%20the%20%27x%27%20property.%22%22%22%0A%20%20%20%20%20%20%20%20print(%22getter%20of%20x%20called%22)%0A%20%20%20%20%20%20%20%20return%20self._x%0A%0A%20%20%20%20%40x.setter%0A%20%20%20%20def%20x(self%2C%20value)%3A%0A%20%20%20%20%20%20%20%20print(%22setter%20of%20x%20called%22)%0A%20%20%20%20%20%20%20%20self._x%20%3D%20value)).

---
### Checklist:
- [ ] Expose state and key of `JumanjiToDMEnvWrapper` as properties
- [ ] Implement setters for these properties
- [ ] Test that one can instantiate a Jumanji environment (e.g. `jumanji.make("Snake-6x6-v0")`), wrap it with `JumanjiToDMEnvWrapper` and then get and set the corresponding `state` and `key` attributes without leading underscores

PR #30, opened 2022-09-23
feat/expose state and key as properties in DMEnvWrapper

louis-csm:
add state and key as properties in DMEnvWrapper class https://github.com/instadeepai/jumanji/issues/27

### DIFF
--- a/jumanji/wrappers.py
+++ b/jumanji/wrappers.py
@@ -36,6 +36,7 @@ from jax import jit, random
 
 from jumanji import specs, tree_utils
 from jumanji.env import Environment
+from jumanji.testing.fakes import FakeState
 from jumanji.types import Action, TimeStep, restart, termination, transition
 
 State = TypeVar("State")
@@ -133,12 +134,12 @@ class JumanjiToDMEnvWrapper(dm_env.Environment):
             env: `Environment`to wrap to a `dm_env.Environment`.
             key: optional key to initialize the `Environment` with.
         """
+        self.state: Union[State, FakeState]
         self._env = env
         if key is None:
-            self._key = random.PRNGKey(0)
+            self.key = random.PRNGKey(0)
         else:
-            self._key = key
-        self._state: State
+            self.key = key
         self._jitted_reset: Callable[[PRNGKey], Tuple[State, TimeStep]] = jit(
             self._env.reset
         )
@@ -217,19 +218,19 @@ class JumanjiToDMEnvWrapper(dm_env.Environment):
 
     @property
     def state(self) -> State:
-        return self._state
+        return self.state
 
     @state.setter
     def state(self, new_state: State) -> None:
-        self._state = new_state
+        self.state = new_state
 
     @property
     def key(self) -> PRNGKey:
-        return self._key
+        return self.key
 
     @key.setter
     def key(self, new_key: PRNGKey) -> None:
-        self._key = new_key
+        self.key = new_key
 
 
 class MultiToSingleWrapper(Wrapper):
--- a/jumanji/wrappers.py
+++ b/jumanji/wrappers.py
@@ -215,6 +215,22 @@ class JumanjiToDMEnvWrapper(dm_env.Environment):
     def unwrapped(self) -> Environment:
         return self._env
 
+    @property
+    def state(self) -> Any:
+        return self._state
+
+    @state.setter
+    def state(self, new_state: Any) -> None:
+        self._state = new_state
+
+    @property
+    def key(self) -> PRNGKey:
+        return self._key
+
+    @key.setter
+    def key(self, new_key: PRNGKey) -> None:
+        self._key = new_key
+
 
 class MultiToSingleWrapper(Wrapper):
     """A wrapper that converts a multi-agent Environment to a single-agent Environment."""
--- a/jumanji/wrappers.py
+++ b/jumanji/wrappers.py
@@ -138,7 +138,7 @@ class JumanjiToDMEnvWrapper(dm_env.Environment):
             self._key = random.PRNGKey(0)
         else:
             self._key = key
-        self._state: Any
+        self._state: State
         self._jitted_reset: Callable[[PRNGKey], Tuple[State, TimeStep]] = jit(
             self._env.reset
         )
@@ -216,11 +216,11 @@ class JumanjiToDMEnvWrapper(dm_env.Environment):
         return self._env
 
     @property
-    def state(self) -> Any:
+    def state(self) -> State:
         return self._state
 
     @state.setter
-    def state(self, new_state: Any) -> None:
+    def state(self, new_state: State) -> None:
         self._state = new_state
 
     @property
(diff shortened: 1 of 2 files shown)


jumanji/wrappers.py
```diff
@@ -215,6 +215,22 @@ def action_spec(self) -> dm_env.specs.Array:
...
+    def state(self, new_state: Any) -> None:
+        self._state = new_state
+
+    @property
+    def key(self) -> PRNGKey:
+        return self._key
+
+    @key.setter
+    def key(self, new_key: PRNGKey) -> None:
+        self._key = new_key
```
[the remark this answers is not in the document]
louis-csm: It is not what has been asked in the issue https://github.com/instadeepai/jumanji/issues/27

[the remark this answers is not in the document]
louis-csm: I can but it is not what has been asked in [the issue](https://github.com/instadeepai/jumanji/issues/27). Sorry I forgot to link the pr to the issue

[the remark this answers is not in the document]
cemlyn007: Sorry, I have discussed this with @clement-bonnet and we agreed to expose these as public attributes because that is a more compact/cleaner solution.

Thank you for your interest and I am looking forward to reviewing your final contribution. 🙂

cemlyn007 (changes requested):
Requesting that the attributes are made public instead of creating getter and setter properties.

jumanji/wrappers.py
```diff
@@ -215,6 +215,22 @@ def action_spec(self) -> dm_env.specs.Array:
     def unwrapped(self) -> Environment:
         return self._env
 
+    @property
+    def state(self) -> Any:
```
clement-bonnet, line 219: Could we use `State` instead of `Any` here?
```suggestion
    def state(self) -> State:
```
Would you mind applying the same to `self._state` in the `__init__`?

jumanji/wrappers.py
```diff
@@ -215,6 +215,22 @@ def action_spec(self) -> dm_env.specs.Array:
...
+        return self._state
+
+    @state.setter
+    def state(self, new_state: Any) -> None:
```
clement-bonnet, line 223: Same here

clement-bonnet:
Could we please implement a test in the `TestJumanjiEnvironmentToDeepMindEnv` class of `wrappers_test.py` for getting and setting the key and state attributes?

clement-bonnet (changes requested):
Nice work!

jumanji/wrappers_test.py
```diff
@@ -228,6 +235,25 @@ def test_jumanji_environment_to_deep_mind_env__unwrapped(
         """Validates unwrapped property of the wrapped environment."""
         assert isinstance(fake_dm_env.unwrapped, Environment)
 
+    def test_jumanji_environment_to_deep_mind_env__key_getter_setter(
+        self, fake_environment: FakeEnvironment)-> None:
+        """validate state getter and setter"""
```
clement-bonnet, line 240: NIT: Could we please update docstrings to have them consistent with others? We try to keep them case sensitive.
```suggestion
        """Validates state getter and setter."""
```

jumanji/wrappers_test.py
```diff
@@ -228,6 +235,25 @@ def test_jumanji_environment_to_deep_mind_env__unwrapped(
...
+        fake_dm_env = JumanjiToDMEnvWrapper(fake_environment, key=key)
+        assert jnp.array_equal(key, fake_dm_env.key)
+        new_key = jax.random.PRNGKey(10)
+        fake_dm_env.key = new_key
+        assert jnp.array_equal(new_key, fake_dm_env.key)
+    
+    def test_jumanji_environment_to_deep_mind_env__state_getter_setter(
+        self, fake_environment: FakeEnvironment, fake_state: FakeState
+        ) -> None:
+        """validate state getter and setter"""
```
clement-bonnet, line 251: Same here
```suggestion
        """Validates state getter and setters"""
```

jumanji/wrappers_test.py
```diff
@@ -228,6 +235,25 @@ def test_jumanji_environment_to_deep_mind_env__unwrapped(
...
+        key = jax.random.PRNGKey(0)
+        fake_dm_env = JumanjiToDMEnvWrapper(fake_environment, key=key)
+        fake_dm_env.state = fake_state
+        assert fake_state == fake_dm_env.state
```
clement-bonnet, line 255: I think we even want to test reference equality here.
```suggestion
        assert fake_state is fake_dm_env.state
```

jumanji/wrappers_test.py
```diff
@@ -21,6 +21,7 @@
 import jax
 import jax.numpy as jnp
 import jax.random as random
+from matplotlib.backend_bases import key_press_handler
```
clement-bonnet, line 24: Why is this needed?

jumanji/wrappers_test.py
```diff
@@ -58,6 +59,12 @@ class MockWrapper(Wrapper[FakeState]):
     return MockWrapper
 
 
+@pytest.fixture
+def fake_state() -> FakeState:
+    return FakeState(key=jax.random.PRNGKey(0),
+                     step=jax.random.randint(jax.random.PRNGKey(0), [1], 0, 100))
+
+
```
clement-bonnet, line 67: If the fake state is used in only one test, it may be preferable not to make it a fixture and define it directly in the corresponding test. Also, I believe the step argument must rather be `step=jax.random.randint(jax.random.PRNGKey(0), (), 0, 100)`.

clement-bonnet: I would also change the key so that it's not the same as the other key defined in the `test_jumanji_environment_to_deep_mind_env__state_getter_setter` test.

clement-bonnet (changes requested):
Thank you for your contribution and for the tests. Just left a few comments.

clement-bonnet:
Thank you for your contribution! Looks good to me :rocket: 

It seems that you have not installed the pre-commit hooks, and hence the changes will probably not pass the linters (black at least). I can also see that tests are not executed on the pipeline for some reason (probably not your fault). Let me look into this!

clement-bonnet (approved):

jumanji/wrappers.py
```diff
@@ -215,6 +215,22 @@ def action_spec(self) -> dm_env.specs.Array:
...
+    def state(self, new_state: State) -> None:
+        self._state = new_state
+
+    @property
+    def key(self) -> PRNGKey:
+        return self._key
+
+    @key.setter
+    def key(self, new_key: PRNGKey) -> None:
+        self._key = new_key
```
cemlyn007, line 232: Why are we using properties instead of attributes here?

cemlyn007 (changes requested):

clement-bonnet:
Is there anybody working on this? Otherwise, I'll be closing the PR for now.

PR #39, opened 2022-10-26
feat: explicit version number and testing

alaterre:


jumanji/registration.py
```diff
@@ -61,10 +65,7 @@ def get_env_id(name: str, version: Optional[int] = None) -> str:
     Returns:
         The environment ID.
     """
-    version = version or 0
-    full_name = name + f"-v{version}"
-
-    return full_name
+    return name + f"-v{version}"
```
clement-bonnet, line 68: I'm wondering. Should we check that `version` is an int here?

jumanji/registration.py
```diff
@@ -123,7 +131,7 @@ def register(
     """Register an environment.
 
     Args:
-        id: environment ID, formatted as `(env_name)[-v(version)]`.
+        id: environment ID, formatted as `(env_name)-v(version)`.
```
clement-bonnet, line 134: NIT
```suggestion
        id: environment ID, formatted as `<env_name>-v<version>`.
```

clement-bonnet (dismissed):
Thank you for the contribution

jumanji/registration.py
alaterre: The docstring says it must be an `int`. If the user doesn't respect the pre-condition then no guarantee. In any case, this function is only used internally and is not exposed to the user.

PR #41, opened 2022-11-04
feat: add render modes in binpacking

coyettev:
This PR is a first attempt to create different rendering modes for jumanji environments. It uses the binpacking environment as an example, but it is meant to be generalized and available to all environments. 

This first draft is a proposition of the interface. It uses the same modes as gym (human and rgb_array).

### DIFF
--- a/jumanji/environments/combinatorial/binpack/env.py
+++ b/jumanji/environments/combinatorial/binpack/env.py
@@ -11,9 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 import itertools
-from typing import Dict, Tuple
+from typing import Dict, Literal, Optional, Tuple, Union
 
 import chex
 import jax
@@ -108,6 +107,9 @@ class BinPack(Environment[State]):
         reward_fn: RewardFn = sparse_linear_reward,
         normalize_dimensions: bool = True,
         debug: bool = False,
+        render_mode: Union[
+            Literal["human", "rgb_array"], env_viewer.RenderMode
+        ] = env_viewer.RenderMode.HUMAN,
     ):
         """Instantiate a BinPack environment.
 
@@ -139,7 +141,9 @@ class BinPack(Environment[State]):
         self.obs_num_ems = obs_num_ems
         self.reward_fn = reward_fn
         self.normalize_dimensions = normalize_dimensions
-        self._env_viewer = env_viewer.BinPackViewer("BinPack")
+        self._env_viewer = env_viewer.BinPackViewer(
+            "BinPack", env_viewer.RenderMode(render_mode)
+        )
         self.debug = debug
 
     def __repr__(self) -> str:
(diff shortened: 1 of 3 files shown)


jumanji/environments/combinatorial/binpack/env.py
```diff
@@ -348,13 +355,13 @@ def step(self, state: State, action: Action) -> Tuple[State, TimeStep[Observatio
 
         return next_state, timestep
 
-    def render(self, state: State) -> None:
+    def render(self, state: State) -> Optional[jnp.ndarray]:
```
cemlyn007, line 358: Good afternoon, I do not think you should narrow the type to `jnp.ndarray` because this cannot handle 3D representations like {'objectId': {'vertices': [...], 'faces': [...], face_colours: [...], etc....}. Have you seen how [Brax](https://github.com/google/brax) does it? I think what they are done is great.

In summary, whilst yes you might want 2D views of the environment, we should also support visualising 3D environments.

coyettev: Good afternoon :slightly_smiling_face:

* Good point for 3d environment ! What would you suggest as return type then ? `Any` ?
* From what I understand in brax, they can either render an html viewer though brax.io.html.render or return an image through brax.io.image.render. The html.render uses some mesh under the hood, but does not return them. I think the mesh representation they use is anyway tightly coupled to their javascript viewer, so it would not make sens to export outside of brax. The interface is similar to what we have now: for a 3d environment we would represent it in whatever visualization framework we have (matplotlib here for binpack), but if we want to export something it would be a 2d image
* We could add a new mode like `RenderMode.MESH`, which would return a mesh representation, but it would have to be a standard mesh type usable by other libraries, do you know if such exists ?

clement-bonnet (commented):
LGTM, @cemlyn007 does this look fine to you?

jumanji/environments/combinatorial/binpack/env_viewer.py
```diff
@@ -12,8 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from enum import Enum
```
cemlyn007, line 15: [Import modules please](https://google.github.io/styleguide/pyguide.html#22-imports).

jumanji/environments/combinatorial/binpack/env_viewer.py
```diff
@@ -24,23 +26,38 @@
 from jumanji.environments.combinatorial.binpack.types import State, item_from_space
 
 
+class RenderMode(str, Enum):
```
cemlyn007, line 29: Why are you inheriting from `str`? Why not like [this](https://docs.python.org/3.8/library/enum.html#creating-an-enum)?

jumanji/environments/combinatorial/binpack/env.py
cemlyn007: Apologies, I misread the PR and thought you wanted to enforce this return type on the Env interface, I am discarding this comment.

jumanji/environments/combinatorial/binpack/env_viewer.py
coyettev: If it inherits from str as well, it can be instantiated from a str parameter, like `render_mode = RenderMode("human")`. As such we can accept either a str or an enum instance in the constructor of the env. Otherwise turning a string to an instance is quite painful

It is kind of like [StrEnum](https://docs.python.org/3/library/enum.html#enum.StrEnum) before python 3.11

jumanji/environments/combinatorial/binpack/env_viewer.py
coyettev: Sure, I always forget this one sorry

Issue #31, opened 2022-10-03
refactor: improve consistency of extras


clement-bonnet:
### Is your feature request related to a problem? Please describe
The `extras` field of `TimeStep` can contain environment information useful for decision-making (e.g. `Connect4`'s current player ID) or environment metrics (e.g. BinPack's volume utilisation). There is an inconsistency in what the `extras` field is used for as it is sometimes meant to be used by the algorithm and sometimes just logged as a metric.

### Describe the solution you'd like
We should move any algorithm-related information from extras to the environment observation (e.g. Connect4's observation could have another field called `current_player` or something). We should update the documentation/docstrings accordingly to explicitly mention that `TimeStep.extras` does not contain stuff that is meant to be observed by the agent as those should be in the observation.

---
### TODOs
- [ ] adapt docstrings, doc, codes, etc to make explicit the fact that `TimeStep.extras` does not contain any info meant to be observed
- [ ] move agent-specific extras (e.g. Connect4's current player ID) to environment observations

clement-bonnet:
It has been discussed that keeping `extras` as naming is better than forcing `metrics`. I am changing the issue description accordingly.

PR #44, opened 2022-11-14
refactor: improve consistency of extras

dluo96:
Closes https://github.com/instadeepai/jumanji/issues/31.

- [x] Adapted docstrings around `extras` to make it clear that `TimeStep.extras` does not contain any information meant to be observed by the agent. Only `observation` does.
- [x] Move agent-specific quantities (for example the `current_player` in the Connect4 environment's `extras` field) from `extras` to the `observation` field.

### DIFF
--- a/jumanji/environments/games/connect4/types.py
+++ b/jumanji/environments/games/connect4/types.py
@@ -14,6 +14,7 @@
 
 from typing import TYPE_CHECKING
 
+import jax.numpy as jnp
 from chex import Array
 from typing_extensions import TypeAlias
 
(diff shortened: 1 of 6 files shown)


jumanji/environments/games/connect4/types.py
```diff
@@ -35,3 +36,4 @@ class State:
 class Observation:
     board: Board
     action_mask: Array
+    current_player: jnp.int8
```
dluo96, line 39: Are we happy to have this be `jnp.int8` as opposed to `int`?

dluo96: Yes.

clement-bonnet (approved):
Thanks for your contribution! :rocket:

PR #46, opened 2022-11-15
fix: resolve jumanji import crash in non-terminal and non-notebook setting

### LINKED ISSUE #34: bug: import jumanji crash from a console that is not a notebook
### Description

If I run `import jumanji` from a console (a PyCharm console in my case), I get the following error from [this line](https://github.com/instadeepai/jumanji/blob/main/jumanji/environments/__init__.py#L44):
```
IPython.core.error.UsageError: Invalid GUI request 'notebook', valid ones are:dict_keys(['none', 'osx', 'tk', 'gtk', 'wx', 'qt', 'qt4', 'qt5', 'glut', 'pyglet', 'gtk3'])
```

I think it comes from the fact that we are checking the backend to set up the proper matplotlib backend accordingly. However, the current version seems to assume a jupyter notebook when it is a python console, thus breaking at import time.

The solution would be to set up the backend optionally, i.e. if an error is encountered, then a default backend is set. Or to figure out how to properly differentiate jupyter notebooks from something else. In any case, we should not break at import time because of rendering!

### What Jumanji version are you using?

v0.1.1

### Which accelerator(s) are you using?

_No response_

### Additional System Info

Linux

### Additional Context

_No response_

### (Optional) Suggestion

_No response_

dluo96:
Closes #34. 

Added another condition in the `environments` folder's `__init__.py` to check if `jumanji` is being run inside a non-Colab notebook. When it is not run in a notebook (Colab or otherwise), the `matplotlib` backend is set to an empty string - `IPython` will deduce the correct choice of backend from inspecting the OS. For example, the backend is inferred to be `MacOSX` for MacOS if we run `IPython.get_ipython().run_line_magic("matplotlib", "")`.

### DIFF
--- a/jumanji/environments/__init__.py
+++ b/jumanji/environments/__init__.py
@@ -40,7 +40,7 @@ try:
         if is_colab():
             backend = "inline"
         else:
-            backend = "notebook"
+            backend = ""
         IPython.get_ipython().run_line_magic("matplotlib", backend)
 
 except ImportError as exc:
--- a/jumanji/environments/__init__.py
+++ b/jumanji/environments/__init__.py
@@ -28,6 +28,10 @@ def is_colab() -> bool:
     return "google.colab" in sys.modules
 
 
+def is_notebook() -> bool:
+    return "ipykernel" in sys.modules
+
+
 # In a notebook, force the Matplotlib backend to ngAgg in order for figures to update
 # every time render is called for environments that use Matplotlib
 # for rendering. Without this, only the last render result is shown per figure.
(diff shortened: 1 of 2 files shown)


jumanji/environments/__init__.py
```diff
@@ -39,8 +43,10 @@ def is_colab() -> bool:
 
         if is_colab():
             backend = "inline"
-        else:
+        elif is_notebook():
```
clement-bonnet, line 46: It seems to me that a PyCharm console has `"ipykernel" in sys.modules` True. Which means the `IPython.get_ipython().run_line_magic("matplotlib", backend)` line would still break on PyCharm console.

dluo96: Have you tested this in the PyCharm Python Console? On my end, `"ipykernel" in sys.modules` returns `False`

clement-bonnet (approved):

PR #52, opened 2022-11-16
fix: import cvrp for environment registration

clement-bonnet:
CVRP was not imported in the jumanji.environments package, leading to `jumanji.make("CVRP20-v0")` failing.


dluo96 (approved):
LGTM!

PR #61, opened 2023-01-03
refactor: binpack instantiation of instance generator

### LINKED ISSUE #60: bug: import of jumanji make pytest tests seg faults.
### Description

Good morning, 

Importing jumanji is making my other tests seg faults. 

Without any test calling/importing code relying on JJ - if I just add a simple import jumanji - tests seg fault, if I remove the import, tests pass.

if I comment in `jumanji/__init__.py` the import of the binpack sub-module (from `jumanji.environments.combinatorial import binpack as _binpack`) and the associated env registrations - tests pass.

I suspect it has to do with something in `jumanji\environments\__init__.py` which is loaded when importing binpack. Since there are a lot of import in the init - it's hard to find the culprit.

I can't disclose the dependencies I am using.

Thanks,

Cyprien

### What Jumanji version are you using?

0.1.3

### Which accelerator(s) are you using?

CPU/GPU

### Additional System Info

Python 3.10.8 - WSL - Ubuntu 20.04 LTS

### Additional Context

_No response_

### (Optional) Suggestion

I have forked the repo and am working on making the registration of the binpack env not require any import from the env. 

Instead of:
```python
register(
    id="BinPack-rand20-v0",
    entry_point="jumanji.environments:BinPack",
    kwargs={
        "instance_generator": _binpack.instance_generator.RandomInstanceGenerator(
            max_num_items=20,
            max_num_ems=80,
        ),
        "obs_num_ems": 40,
    },
)
```

We could have:
```python
register(
    id="BinPack-rand20-v0",
    entry_point="jumanji.environments:BinPack",
    kwargs={
        "instance_generator": {
            "type": "random",
            "max_num_items": 20,
            "max_num_ems": 80,
        },
        "obs_num_ems": 40,
    },
)
```
or,
```python
register(
    id="BinPack-rand20-v0",
    entry_point="jumanji.environments:BinPack",
    kwargs={
        "instance_generator": "random",
        "max_num_items": 20,
        "max_num_ems": 80,
        "obs_num_ems": 40,
    },
)
```

cyprienc:
Resolves #60 

This moves the instantiation of the instance generator for the binpack environment to the init of the binpack env.

Previously - this was done at registration.

### DIFF
--- a/jumanji/environments/combinatorial/binpack/env.py
+++ b/jumanji/environments/combinatorial/binpack/env.py
@@ -105,7 +105,7 @@ class BinPack(Environment[State]):
 
     def __init__(
         self,
-        instance_generator: str = "toy",
+        instance_generator_type: str = "toy",
         obs_num_ems: int = 60,
         reward_fn: RewardFn = sparse_linear_reward,
         normalize_dimensions: bool = True,
@@ -115,10 +115,10 @@ class BinPack(Environment[State]):
         """Instantiate a BinPack environment.
 
         Args:
-            instance_generator: string representing the InstanceGenerator responsible for resetting
-                the environment. E.g. can be a random generator to learn generalisation or one that
-                outputs the same instance to do active search on that instance.
-                Default to ToyInstanceGenerator that always resets to the same instance
+            instance_generator_type: string representing the InstanceGenerator responsible for
+                resetting the environment. E.g. can be a random generator to learn generalisation
+                or one that outputs the same instance to do active search on that instance.
+                Defaults to ToyInstanceGenerator that always resets to the same instance
                 with 20 items. Possible values: 'toy' (default), 'csv' or 'random'.
             obs_num_ems: number of ems to show to the agent. If `obs_num_ems` is smaller than
                 `generator.max_num_ems`, the first `obs_num_ems` biggest ems will be returned
@@ -136,7 +136,7 @@ class BinPack(Environment[State]):
             instance_generator_kwargs: Keyword arguments for the specified instance generator.
         """
         self.instance_generator = self.create_instance_generator(
-            instance_generator, **instance_generator_kwargs
+            instance_generator_type, **instance_generator_kwargs
         )
 
         self.obs_num_ems = obs_num_ems
@@ -149,7 +149,7 @@ class BinPack(Environment[State]):
         return "\n".join(
             [
                 "BinPack environment:",
-                f" - instance_generator: {self.instance_generator}",
+                f" - instance_generator_type: {self.instance_generator}",
                 f" - max_num_items: {self.instance_generator.max_num_items}",
                 f" - obs_num_ems: {self.obs_num_ems}",
                 f" - max_num_ems: {self.instance_generator.max_num_ems}",
@@ -161,7 +161,7 @@ class BinPack(Environment[State]):
 
     @classmethod
     def create_instance_generator(
-        cls, instance_generator: str, **instance_generator_kwargs: Any
+        cls, instance_generator_type: str, **instance_generator_kwargs: Any
     ) -> InstanceGenerator:
         """
         Factory method for creating an instance generator.
@@ -169,7 +169,7 @@ class BinPack(Environment[State]):
         This method can be overridden to add new instance generator types.
 
         Args:
-            instance_generator: The type of instance generator to create. Possible values:
+            instance_generator_type: The type of instance generator to create. Possible values:
                 - 'toy': Create a toy instance generator.
                 - 'csv': Create a CSV instance generator.
                 - 'random': Create a random instance generator.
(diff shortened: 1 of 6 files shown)


jumanji/environments/combinatorial/binpack/env.py
```diff
@@ -156,6 +159,46 @@ def __repr__(self) -> str:
...
+        elif instance_generator == "random":
+            instance_generator_obj = RandomInstanceGenerator(
+                **instance_generator_kwargs
+            )
+        else:
+            raise ValueError(
+                f"Unexpected value for 'instance_generator', got {instance_generator!r}."
+                "Possible values: 'toy', 'csv', 'random'."
+            )
+        return instance_generator_obj
```
clement-bonnet, line 200: Small suggestion to factorise this a little (not sure about first-line typing).
```suggestion
        instance_generator_cls: Type[InstanceGenerator]

        if instance_generator == "toy":
            instance_generator_cls = ToyInstanceGenerator
        elif instance_generator == "csv":
            instance_generator_cls = CSVInstanceGenerator
        elif instance_generator == "random":
            instance_generator_cls = RandomInstanceGenerator
        else:
            raise ValueError(
                f"Unexpected value for 'instance_generator', got {instance_generator!r}."
                "Possible values: 'toy', 'csv', 'random'."
            )
        return instance_generator_cls(**instance_generator_kwargs)
```

Since this is essentially a factory method, we could alternatively use a dictionary property
```
self.instance_generator_factory = {
    "toy": ToyInstanceGenerator,
    "csv": CSVInstanceGenerator,
    "random": RandomInstanceGenerator,
}
```
But this is essentially the same, what do you think?

jumanji/environments/combinatorial/binpack/env.py
```diff
@@ -103,20 +105,21 @@ class BinPack(Environment[State]):
 
     def __init__(
         self,
-        instance_generator: InstanceGenerator = None,
+        instance_generator: str = "toy",
         obs_num_ems: int = 60,
         reward_fn: RewardFn = sparse_linear_reward,
```
clement-bonnet, line 110: Outside of the scope of this PR, we could use a similar factory for the reward function to avoid passing complex objects to the constructor. Having `int`, `float` and `str` as constructor arguments facilitate a lot.

clement-bonnet (commented):
Thank you for your contribution. Making the instance generator argument a str removes some issues we have been facing and will be a preferred choice going forward.

jumanji/environments/combinatorial/binpack/env.py
```diff
@@ -103,20 +105,21 @@ class BinPack(Environment[State]):
...
     ):
         """Instantiate a BinPack environment.
 
         Args:
-            instance_generator: InstanceGenerator responsible for resetting the environment. E.g.
-                can be a random generator to learn generalisation or one that outputs the same
-                instance to do active search on that instance. It must inherit from the
-                InstanceGenerator abstract class. Default to ToyInstanceGenerator that always
-                resets to the same instance with 20 items.
+            instance_generator: string representing the InstanceGenerator responsible for resetting
```
djbyrne, line 118: can we rename this to something like "instance_generator_type" ? as it is no longer the instance generator but a flag

jumanji/environments/combinatorial/binpack/env.py
cyprienc: Thanks for the suggestions. 

The `ToyInstanceGenerator` class doesn't take any parameter so linters won't allow to pass (even empty) kwargs because of that.

Regarding the dictionary vs class method -  I lean towards the class method one. It's closer to the "strategy" design pattern, uses the OOP principles, is a bit more explicit and 'immutable'.

jumanji/environments/combinatorial/binpack/env.py
cyprienc: I'll let you create an (internal?) issue and mark this as resolved.

jumanji/environments/combinatorial/binpack/env.py
cyprienc: done.

jumanji/environments/combinatorial/binpack/env.py
clement-bonnet: I see, thanks!
@djbyrne what do you think is preferred in this case? A more factorised code and a type ignore or the initial solution?

djbyrne: TBH, I am not a fan of type ignores if we can avoid them. So only based on that I would lean toward just instantiating the objects using the initial solution. 

@clement-bonnet  I do still like the idea of having a directory of classes that is used to instantiate the object. I think it is still following the factory design pattern as it is creational as opposed to operational and would argue that it is still following OOP principles if it wasn't for the linting error.

djbyrne (dismissed):

djbyrne (dismissed):

clement-bonnet (dismissed):

clement-bonnet (approved):

PR #68, opened 2023-02-20
feat: increment semantic release

clement-bonnet:
- Had to install Brax from source as any version after 0.0.16 (and potentially slightly before) is broken with recent versions of Jax. We will probably remove Brax from Jumanji's dependencies in the future.


dluo96 (dismissed):
LGTM

dluo96 (approved):
LGTM

PR #69, opened 2023-02-21
refactor: remove brax

clement-bonnet:
Packages installed from PyPI cannot depend on packages which are not also hosted on PyPI. Brax releases are broken with the newest version of Jax. Brax is not needed in Jumanji and should be removed from the core dependencies. I see very little transfer with agent code from Brax to Jumanji as Brax has quite a few brax-specific features (normalization, truncation, etc), hence weakening the purpose of this wrapper. I suggest removing Brax entirely.

Same question arises for Gymnax.


dluo96 (approved):
LGTM

PR #67, opened 2023-02-15
feat: deprecate Connect4

clement-bonnet:
- deprecate Connect4 by throwing a warning
- fix Connect4 action mask type
- downgrade pygame to support python 3.7

PR #72, opened 2023-03-10
fix: pygame requirements

wang-r-j:
The pygame compatibility issue with python 3.7 first appeared in pygame 2.1.3.dev6 (pygame/pygame@a462721e) and has been fully resolved in 2.2.0.dev2 (pygame/pygame#3723, pygame/pygame#3572, pygame/pygame@aab517c7). Therefore, the requirement tightened in #67 can be relaxed and only pygame 2.1.3.* needs to be excluded.


clement-bonnet (approved):
Thank you for looking into this and for the fix!

PR #99, opened 2023-03-28
feat: create viewer interface for rendering of environments

### LINKED ISSUE #93: refactor: create env viewer interface
Create an abstract base class `Viewer` which all environment viewers inherit from. This will help standardise how rendering is done across all environments.

dluo96:
Closes #93. 

Created an environment viewer interface that all environment viewers should inherit from. This will allow users to extend the `Viewer` class to implement their own rendering. This viewer can be passed directly to the init of the environment of interest.


jumanji/env_viewer.py
```diff
@@ -0,0 +1,62 @@
...
+
+from jumanji.env import State
+
+
+class Viewer(abc.ABC, Generic[State]):
+    """Abstract viewer class to support rendering and animation. This interface assumes
+    that matplotlib is used for rendering the environment in question.
+    """
+
+    def render(self, state: State) -> Optional[NDArray]:
```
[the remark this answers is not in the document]
clement-bonnet: Either nothing (if printed to the screen) or a numpy array (if one wants to manipulate the array of pixels).

jumanji/env_viewer.py
```diff
@@ -0,0 +1,62 @@
...
+import abc
+from typing import Generic, Optional, Sequence
+
+import matplotlib
+from numpy.typing import NDArray
+
+from jumanji.env import State
+
+
+class Viewer(abc.ABC, Generic[State]):
```
clement-bonnet, line 26: I wonder, if we assume matplotlib, should we call this `MatplotlibViewer` or something?

jumanji/env_viewer.py
clement-bonnet, line 31: Optional: we could use `abstractmethod` instead of `raise NotImplementedError`. This would be more consistent with the `Environment` abstract class.
```suggestion
    @abc.abstractmethod
    def render(self, state: State) -> Optional[NDArray]:
```

jumanji/env_viewer.py
```diff
@@ -0,0 +1,62 @@
...
+    that matplotlib is used for rendering the environment in question.
+    """
+
+    def render(self, state: State) -> Optional[NDArray]:
+        """Render frames of the environment for a given state using matplotlib.
+
+        Args:
+            state: `State` object corresponding to the new state of the environment.
+        """
+        raise NotImplementedError
```
clement-bonnet, line 37: We can remove this if we go for the above
```suggestion
```

jumanji/env_viewer.py
```diff
@@ -0,0 +1,62 @@
...
+        self,
+        states: Sequence[State],
+        interval: int,
+        save_path: Optional[str],
+    ) -> matplotlib.animation.FuncAnimation:
+        """Create an animation from a sequence of environment states.
+
+        Args:
+            states: sequence of environment states corresponding to consecutive timesteps.
+            interval: delay between frames in milliseconds, default to 200.
```
clement-bonnet, line 49: No default here
```suggestion
            interval: delay between frames in milliseconds.
```

jumanji/environments/packing/knapsack/env.py
```diff
@@ -96,14 +98,16 @@ def __init__(
                 transition. The function must compute the reward based on the current state,
                 the chosen action, the next state and whether the action is valid.
                 Implemented options are [`DenseReward`, `SparseReward`]. Defaults to `DenseReward`.
+            viewer: `Viewer` used for rendering. Defaults to `KnapsackViewer`.
             render_mode: string that defines the mode of rendering.
```
clement-bonnet, line 102: We probably should not have the choice of the render mode here, it seems to be mixing composite and strategy pattern. We just want the `env_viewer` that defaults to some `render_mode` right?

jumanji/environments/routing/cleaner/env.py
```diff
@@ -95,7 +97,9 @@ def __init__(
             generator: `Generator` whose `__call__` instantiates an environment instance.
                 Implemented options are [`RandomGenerator`]. Defaults to `RandomGenerator` with
                 `num_rows=10` and `num_cols=10`.
+            viewer: `Viewer` used for rendering. Defaults to `CleanerViewer`.
             render_mode: the mode for visualising the environment, can be "human" or "rgb_array".
```
clement-bonnet, line 101: Same comment regarding render_mode for all the other envs

clement-bonnet (changes requested):
Thanks for the PR, looks much better like this! Three points:
- Should we rename the abstract viewer to account for the fact that it assumes matplotlib?
- Should we move render, close and animate to the `Environment` parent class and call super(viewer) in each environment?
- Should we stay compositional and not have render_mode as another argument alongside viewer?

clement-bonnet:
After discussion, updating the comment.

> Thanks for the PR, looks much better like this! Three points:
> 
> * Should we rename the abstract viewer to account for the fact that it assumes matplotlib?

Actually, it's fine, we can just state we use matplotlib for it in the docstring.

> * Should we move render, close and animate to the `Environment` parent class and call super(viewer) in each environment?

It's better not to do so to make it clear what methods are implemented on each environment.

> * Should we stay compositional and not have render_mode as another argument alongside viewer?

Indeed, we should just have
```
def __init__(self, ..., viewer: Optional[Viewer[State]]):
    ...
    self._viewer = viewer or DefaultViewer(..., render_mode=...)
```

jumanji/env_viewer.py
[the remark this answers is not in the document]
TristanKalloniatis: Do any of them do the latter?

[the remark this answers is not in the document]
dluo96: Some of the environments do

jumanji/env_viewer.py
dluo96: Agreed to stick to `Viewer`

clement-bonnet (approved):
LGTM

PR #105, opened 2023-03-28
Small typo terminal logger

clement-bonnet:
- fix typo in terminal logger
- augment the BinPack doc to clarify that the implemented version is equivalent to 3D-knapsack and not multi-bin packing @Rebecca


dluo96 (approved):
LGTM 🚀

PR #122, opened 2023-04-26
feat(maze): update generator to return state

PDuckworth:
Updated the generator to return `State`.

### DIFF
--- a/jumanji/environments/routing/maze/generator.py
+++ b/jumanji/environments/routing/maze/generator.py
@@ -14,9 +14,11 @@
 import abc
 
 import chex
+import jax
 import jax.numpy as jnp
 
 from jumanji.environments.commons.maze_utils import maze_generation
+from jumanji.environments.routing.maze.types import Position, State
 
 
 class Generator(abc.ABC):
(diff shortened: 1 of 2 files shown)


jumanji/environments/routing/maze/generator.py
```diff
@@ -72,8 +86,35 @@ def __call__(self, key: chex.PRNGKey) -> chex.Array:
...
-            maze: A generated maze as an array of booleans.
+            state: the generated state.
         """
-        return maze_generation.generate_maze(self.num_cols, self.num_rows, key).astype(
-            bool
+        key, maze_key, agent_key = jax.random.split(key, 3)
+
+        walls = maze_generation.generate_maze(
+            self.num_cols, self.num_rows, maze_key
+        ).astype(bool)
```
dluo96, line 95: Are the first two arguments to this function in the correct order? I ask because `generate_maze` states:

```
def generate_maze(width: int, height: int, key: chex.PRNGKey) -> chex.Array:
    """Randomly generate a maze.

    Args:
        width: the number of rows of the maze to create.
        height: the number of columns of the maze to create.
        key: the Jax random number generation key.
```
so it seems that the first argument should be the number of rows while the second should be the number of columns?

dluo96 (changes requested):
Thank you for your contribution @PDuckworth! I have left one comment, otherwise it looks good to me! 🚀

jumanji/environments/routing/maze/generator.py
PDuckworth: Hi @dluo96 - I think the arguments in `maze_generation.generate_maze(....)` are in the correct order. But I think the `generate_maze` doc string is incorrect. The `width` of a maze is the number of columns (not rows) the maze has, and the `height` should be the number of rows (not cols).

PDuckworth: Good spot though!

clement-bonnet (approved):
Thanks for the contribution, Paul!

jumanji/environments/routing/maze/generator.py
PDuckworth: Thanks @dluo96! 
I believe this now closes Issue https://github.com/instadeepai/jumanji/issues/131

Issue #106, opened 2023-03-29
fix: step type in autoreset wrapper


dluo96:
### Is your feature request related to a problem? Please describe
When using `AutoResetWrapper`, the sequence of `timestep.step_type` returned during a rollout shows 0 values where the env has been auto reset instead of 2 values. Recall that

- 0 corresponds to the first step
- 1 corresponds to a mid step
- 2 corresponds to the last step

### Describe the solution you'd like
Move the auto reset bug: `[1, 1, 1, 2, 0, 1, 1, 1]` has to be converted to `[1, 1, 1, X, 1, 1, 1]` with X being 0 or 2. Right now it is 0 but it should be 2 to warn the user that the episode got terminated. This would be important e.g. if using `while not timestep.last()`. 

```
timestep = timestep.replace(  # type: ignore
    observation=reset_timestep.observation
)
```

clement-bonnet:
Got fixed in #123

PR #123, opened 2023-04-26
fix: make autoreset wrapper return 2 on reset

rodSiry:
Fixed the bad autoreset behavior described in issue #106 

Now, wrapped environment returns step_type = 2 when resetting.

Adapted corresponding test function, with assert step_type == 2.

### DIFF
--- a/jumanji/wrappers.py
+++ b/jumanji/wrappers.py
@@ -380,8 +380,7 @@ class AutoResetWrapper(Wrapper):
 
         # Replace observation with reset observation.
         timestep = timestep.replace(  # type: ignore
-            observation=reset_timestep.observation,
-            step_type=reset_timestep.step_type,
+            observation=reset_timestep.observation
         )
 
         return state, timestep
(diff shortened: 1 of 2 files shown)


dluo96 (changes requested):
Thank you for your contribution @rodSiry! 🚀  This looks good to me! Could you update the PR title to follow the conventional commit style format (see our [contribution guidelines](https://github.com/instadeepai/jumanji/blob/main/CONTRIBUTING.md#coding-style)) and to be more informative? For example: `fix: make autoreset wrapper return 2 on reset`.

dluo96 (dismissed):

clement-bonnet:
Hi, thanks for the contribution. Would you mind applying the same fix to the `VmapAutoResetWrapper`? By saying so, I realize we should have shared some code between the two...
Thanks again!

clement-bonnet (changes requested):
Hi, thanks for the contribution. Would you mind applying the same fix to the `VmapAutoResetWrapper`? By saying so, I realize we should have shared some code between the two...
Thanks again!

clement-bonnet (changes requested):
Looks good to me! One last thing I noticed: we need to update the class docstring of `AutoResetWrapper` to account for not resetting the step_type to FIRST anymore. Would you mind doing so? Thanks!

clement-bonnet (approved):
Thanks for the fix!

PR #130, opened 2023-05-08
feat(graph_coloring): implement graph_coloring environment

ElshadaiK:
Description:

This pull request introduces a new graph coloring network designed to solve the single-agent graph coloring problem. The network uses a combination of node and color embeddings, self-attention, and cross-attention mechanisms to generate high-quality color assignments for graph nodes. The key components of this PR include:

A graph coloring environment that represents the problem and provides necessary information for the agent to learn the optimal coloring.
An agent network that combines node and color embeddings, self-attention, and cross-attention mechanisms to make intelligent decisions on color assignments.
Transformer-based architecture with multiple layers, heads, and MLP units for improved learning and adaptability.


jumanji/training/networks/graph_coloring/actor_critic.py
```diff
@@ -0,0 +1,283 @@
...
+        self.model_size = transformer_num_heads * transformer_key_size
+
+    def _make_self_attention_mask(self, adj_matrix: chex.Array) -> chex.Array:
+        # Expand on the head dimension.
+        mask = jnp.expand_dims(adj_matrix, axis=-3)
+        return mask
+
+    def embed_nodes(self, nodes: chex.Array) -> chex.Array:
+        nodes_float = nodes.astype(jnp.float32)
+        nodes_leaves = jnp.stack(jtree.tree_leaves(nodes_float), axis=-1)
```
clement-bonnet, line 90: Same comment here as below.
```suggestion
        nodes_leaves = nodes_float
```

jumanji/training/networks/graph_coloring/actor_critic.py
```diff
@@ -0,0 +1,283 @@
...
+                node_embeddings,
+                colors_cross_nodes_mask,
+            )
+
+            node_embeddings = new_node_embeddings
+
+        batch_indices = jnp.arange(node_embeddings.shape[0])
+        current_node_embeddings = node_embeddings[
+            batch_indices, observation.current_node_index
+        ]
```
clement-bonnet, line 223: LGTM. Alternatively, you can use `jnp.take` or `jnp.take_along_axis`.

jumanji/training/networks/graph_coloring/actor_critic.py
```diff
@@ -0,0 +1,283 @@
...
+    def network_fn(observation: Observation) -> chex.Array:
+        torso = GraphColoringTorso(
+            num_transformer_layers=num_transformer_layers,
+            transformer_num_heads=transformer_num_heads,
+            transformer_key_size=transformer_key_size,
+            transformer_mlp_units=transformer_mlp_units,
+            name="policy_torso",
+        )
+        embeddings = torso(observation)  # (B, N, H)
+        logits = hk.Linear(1, name="policy_head")(embeddings)  # (B, N, 1)
```
clement-bonnet, line 252: Why have a linear layer here versus an MLP (2 layers) for the critic below?

clement-bonnet (changes requested):
Thank you so much for your contribution. The code looks very good. I have a few comments on small things. Please let me know how you feel about them.

clement-bonnet (changes requested):
Thank you so much for your contribution. The code looks very good. I have a few comments on small things. Please let me know how you feel about them.

jumanji/training/networks/graph_coloring/actor_critic.py
```diff
@@ -0,0 +1,283 @@
...
+            colors_array == color_indices
+        )  # Shape (batch_size, num_colors, num_nodes)
+
+        # Expand and repeat along the transformer_num_heads axis
+        colors_cross_nodes_mask = jnp.expand_dims(
+            colors_cross_nodes_mask, axis=1
+        )  # Shape (batch_size, 1, num_colors, num_nodes)
+        colors_cross_nodes_mask = jnp.repeat(
+            colors_cross_nodes_mask, repeats=self.transformer_num_heads, axis=1
+        )  # Shape (batch_size, transformer_num_heads, num_colors, num_nodes)
```
clement-bonnet, line 144: Actually, this should not be necessary, as the mask is automatically repeated on the `num_heads` dimension.
```suggestion
```

jumanji/training/networks/graph_coloring/actor_critic.py
```diff
@@ -0,0 +1,283 @@
...
+            colors_array == color_indices
+        )  # Shape (batch_size, num_nodes, num_colors)
+
+        # Expand and repeat along the transformer_num_heads axis
+        nodes_cross_colors_mask = jnp.expand_dims(
+            nodes_cross_colors_mask, axis=1
+        )  # Shape (batch_size, 1, num_nodes, num_colors)
+        nodes_cross_colors_mask = jnp.repeat(
+            nodes_cross_colors_mask, repeats=self.transformer_num_heads, axis=1
+        )  # Shape (batch_size, transformer_num_heads, num_nodes, num_colors)
```
clement-bonnet, line 165: Same here
```suggestion
```

jumanji/training/networks/graph_coloring/actor_critic.py
```diff
@@ -0,0 +1,283 @@
...
+        colors_array = jnp.expand_dims(
+            observation.colors, axis=-1
+        )  # Shape (batch_size, num_nodes, 1)
+        color_indices = jnp.arange(observation.colors.shape[1])  # Shape (num_colors,)
+        color_indices = jnp.expand_dims(color_indices, axis=0)  # Shape (1, num_colors)
+        color_indices = jnp.expand_dims(
+            color_indices, axis=0
+        )  # Shape (1, 1, num_colors)
+
+        nodes_cross_colors_mask = (
```
clement-bonnet, line 155: It seems to me here that `nodes_cross_colors_mask` is equal to `colors_cross_nodes_mask`, but they should be a transpose of each other, right?

jumanji/environments/logic/graph_coloring/env.py
```diff
@@ -0,0 +1,264 @@
...
+                and `percent_connected = 0.8` parameters.
+        """
+        self.generator = generator or RandomGenerator(
+            num_nodes=100,
+            percent_connected=0.8,
+        )
+        num_nodes, percent_connected = self.generator.specs()
+        # Create viewer used for rendering
+        self._env_viewer = GraphColoringViewer(
+            num_nodes=num_nodes, name="Graph Coloring"
```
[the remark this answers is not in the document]
coyettev: From the rest of the code it seems that the `num_nodes` is assumed to be constant, it is totally fine, but I still think the viewer should not depend on it. in what @clement-bonnet suggested (possibility to have the viewer as an argument and not create it here), we wouldn't know in advance the num_nodes

jumanji/environments/logic/graph_coloring/viewer.py
```diff
@@ -0,0 +1,211 @@
...
+
+        for i in range(self._num_nodes):
+            for j in range(i + 1, self._num_nodes):
+                if state.adj_matrix[i, j]:
+                    ax.plot(
+                        [pos[i][0], pos[j][0]],
+                        [pos[i][1], pos[j][1]],
+                        color=self._color_mapping[i],
+                        linewidth=0.5,
+                    )
```
coyettev, line 84: Maybe this function ca be separated in smaller functions so that it is easier to follow ? 
```python 
    def render(
        self,
        state: State,
        save_path: Optional[str] = None,
        ax: Optional[plt.Axes] = None,
    ) -> None:
        self._clear_display()
        fig, ax = self._get_fig_ax(ax)
        self._render_nodes(ax, state)
        self._render_edges(ax, state)
...
```

jumanji/environments/logic/graph_coloring/viewer.py
```diff
@@ -0,0 +1,211 @@
...
+    def render(
+        self,
+        state: State,
+        save_path: Optional[str] = None,
+        ax: Optional[plt.Axes] = None,
+    ) -> None:
+        self._clear_display()
+
+        node_scale = 5 + int(np.sqrt(self._num_nodes))
+        node_radius = 0.05 * 5 / node_scale
```
coyettev, line 55: Maybe a little comment explaining where this comes from ?

coyettev (changes requested):
Nice environment, congrats ! I left some comments :)

jumanji/environments/logic/graph_coloring/env.py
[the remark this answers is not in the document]
clement-bonnet: Agree with @coyettev that the viewer should not have to know `num_nodes` in advance. As for the environment, it does for now assume a constant `num_nodes`, indeed. This could be augmented by having a dynamic number of nodes and a mask that tells you which nodes exist. Yet, it is not foreseen for this version. If one wants to work with different number of nodes, one can instantiate multiple environments (cf my work on curriculum learning).

jumanji/environments/logic/graph_coloring/env.py
```diff
@@ -0,0 +1,264 @@
...
+        valid_actions = self._get_valid_actions(state)
+        # Check if the chosen action is invalid (not in valid_actions).
+        invalid_action_taken = jnp.logical_not(valid_actions[action])
+
+        # Update the colors array with the chosen action.
+        colors = state.colors.at[state.current_node_index].set(action)
+        colored_nodes = colors >= 0
+
+        # Determine if all nodes have been assigned a color
+        all_nodes_colored = jnp.count_nonzero(colored_nodes) == num_nodes
```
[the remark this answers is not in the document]
clement-bonnet: The latter assumes that `state.current_node_index` always starts at 0, which may not be the case in the future (if a generator decides otherwise). I agree with the former though!

jumanji/environments/logic/graph_coloring/env.py
```diff
@@ -0,0 +1,264 @@
...
+        unique_colors_used = jnp.unique(colors, size=num_nodes, fill_value=-1)
+        num_unique_colors = jnp.count_nonzero(unique_colors_used >= 0)
+        reward = jnp.where(all_nodes_colored, -num_unique_colors, 0.0).sum()
+
+        # Apply the maximum penalty when an invalid action is taken and terminate the episode
+        reward = jnp.where(invalid_action_taken, -num_nodes, reward)
+        done = jnp.logical_or(all_nodes_colored, invalid_action_taken)
+
+        # Update the current node index
+        next_node_index = (state.current_node_index + jnp.array(1)) % num_nodes
```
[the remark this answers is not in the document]
clement-bonnet: Since it will be a terminal state, I don't think it matters what the state/observation is.

docs/environments/graph_coloring 2.md
```diff
@@ -0,0 +1,52 @@
...
+
+The reward in the Graph Coloring environment is given as follows:
+
+- `reward`: A reward is provided based on the number of unique colors used if all nodes have been assigned a color. The reward is equal to the negative of the number of unique colors used.
+
+The agent's goal is to find a valid coloring using as few colors as possible while avoiding conflicts with adjacent nodes.
+
+## Episode Termination
+
+An episode terminates when all nodes in the graph have been assigned a color. The goal of the agent is to find a valid coloring using as few colors as possible.
```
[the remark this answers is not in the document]
ElshadaiK: yes, you're right. Let me adjust it :)

docs/environments/graph_coloring 2.md
```diff
@@ -0,0 +1,52 @@
...
+        [ True, False,  True, False],
+        [False,  True, False,  True],
+        [ True, False,  True, False]]```
+
+- `colors`: jax array (int32) of shape (num_nodes,), representing the current color assignments for the vertices. Initially, all elements are set to -1, indicating that no colors have been assigned yet.
+  - For example, an initial color assignment:
+    ```[-1, -1, -1, -1]```
+
+- `action_mask`: jax array (bool) of shape (num_colors,), indicating which actions are valid in the current state of the environment. The valid actions are represented by True, while invalid actions are represented by False.
+  - For example, for 4 number of colors available:
```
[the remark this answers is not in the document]
ElshadaiK: I have adjusted it. Let me know if it still isn't clear

jumanji/environments/logic/graph_coloring/env.py
```diff
@@ -0,0 +1,264 @@
...
+        next_state = State(
+            adj_matrix=state.adj_matrix,
+            colors=colors,
+            current_node_index=next_node_index,
+            key=state.key,
+        )
+        obs = Observation(
+            adj_matrix=state.adj_matrix,
+            colors=colors,
+            action_mask=self._get_valid_actions(next_state),
```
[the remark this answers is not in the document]
ElshadaiK: hmm okay that makes sense. 

we first need to get the action_mask in the step fn, to check the node isn't colored the color out of the action mask. And in this case it's going to be the action mask of the current node. When we instantiate Observation after stepping, we're getting the action mask for the next node to be colored. But if we store that in the State too, we can get it later when stepping on the new node

jumanji/training/networks/graph_coloring/actor_critic.py
ElshadaiK: I wouldn't say they are the transpose of each other. 

`colors_cross_nodes_mask` represents for each color, which nodes have that color. It has shape (batch_size, transformer_num_heads, num_colors, num_nodes).

`nodes_cross_colors_mask` represents for each node, which colors it has. It has shape (batch_size, transformer_num_heads, num_nodes, num_colors).

jumanji/environments/logic/graph_coloring/env.py
```diff
@@ -0,0 +1,264 @@
...
+        colors = state.colors.at[state.current_node_index].set(action)
+        colored_nodes = colors >= 0
+
+        # Determine if all nodes have been assigned a color
+        all_nodes_colored = jnp.count_nonzero(colored_nodes) == num_nodes
+
+        # Calculate the reward
+        unique_colors_used = jnp.unique(colors, size=num_nodes, fill_value=-1)
+        num_unique_colors = jnp.count_nonzero(unique_colors_used >= 0)
+        reward = jnp.where(all_nodes_colored, -num_unique_colors, 0.0).sum()
```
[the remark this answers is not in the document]
ElshadaiK: what about 

        ```
all_nodes_colored = jnp.all(colors >= 0)

        # Calculate the reward
        def compute_reward(colors):
            unique_colors_used = jnp.unique(colors, size=num_nodes, fill_value=-1)
            num_unique_colors = jnp.count_nonzero(unique_colors_used >= 0)
            return -num_unique_colors

        # Use nested cond for multi-branch condition
        reward = jax.lax.cond(
            invalid_action_taken, 
            lambda _: -num_nodes, 
            lambda _: jax.lax.cond(all_nodes_colored, compute_reward, lambda _: 0, colors),
            None
        )
```

File: jumanji/environments/logic/graph_coloring/env.py [hunk 16]
```diff
@@ -0,0 +1,264 @@
...
+
+        # Apply the maximum penalty when an invalid action is taken and terminate the episode
+        reward = jnp.where(invalid_action_taken, -num_nodes, reward)
```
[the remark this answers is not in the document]
ElshadaiK: what do you suggest if I use another nested function like this? @clement-bonnet @coyettev  

```
         all_nodes_colored = jnp.all(colors >= 0)

        # Calculate the reward
        def compute_reward(colors):
            unique_colors_used = jnp.unique(colors, size=num_nodes, fill_value=-1)
            num_unique_colors = jnp.count_nonzero(unique_colors_used >= 0)
            return -num_unique_colors

        # Use nested cond for multi-branch condition
        reward = jax.lax.cond(
            invalid_action_taken, 
            lambda _: -num_nodes, 
            lambda _: jax.lax.cond(all_nodes_colored, compute_reward, lambda _: 0, colors),
            None
        )
```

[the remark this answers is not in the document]
coyettev: Actually @clement-bonnet is right in its other comment, as the step will generally be vmap-ed, all cond will be converted to select anyway and all branches will be computed, so you can very well keep it as it is :+1:

jumanji/environments/logic/graph_coloring/viewer.py
```diff
@@ -0,0 +1,223 @@
...
+            fig = ax.figure
+            ax.clear()
+        return fig, ax
+
+    def _render_nodes(self, ax: plt.Axes, state: State) -> None:
+        # Set the radius of the nodes as a fraction of the scale,
+        # so nodes appear smaller when there are more of them.
+        node_radius = 0.05 * 5 / self.node_scale
+
+        pos = self._spring_layout(state.adj_matrix)
```
coyettev, line 202: Sorry it is my fault for misleading you, by computing the `pos` here and in `_render_edges` the layout has to be computed twice. Computation is not very critical for the render, but I am more worried by the fact that the spring_layout is stochastic, and I don't think it is very robust to rely on the seed to ensure that `pos` is the same in `_render_nodes` and `_render_edges`.

Maybe `pos` can be an argument to both functions ? or stored in `self.pos` ?

jumanji/training/networks/graph_coloring/actor_critic.py
```diff
@@ -0,0 +1,283 @@
...
+
+    def embed_nodes(self, nodes: chex.Array) -> chex.Array:
+        nodes_float = nodes.astype(jnp.float32)
+        nodes_leaves = jnp.stack(jtree.tree_leaves(nodes_float), axis=-1)
+        embeddings = hk.Linear(self.model_size, name="node_projection")(nodes_leaves)
+        return embeddings
+
+    def embed_colors(self, colors: chex.Array) -> chex.Array:
+        color_float = colors.astype(jnp.float32)
+        colors_leaves = jnp.stack(jtree.tree_leaves(color_float), axis=-1)
```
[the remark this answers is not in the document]
ElshadaiK: The issue here is I need my embedding to be of shape (batch_size, num_nodes, 128), but when I do `colors_leaves = jnp.stack(color_float, axis=-1)`, I get an embedding of shape (num_nodes, 128). Or when doing `embeddings = hk.Linear(self.model_size, name="color_projection")(color_float)`, the embedding will be of shape (batch_size, 128)

jumanji/environments/logic/graph_coloring/viewer.py
ElshadaiK: yeah both work. @clement-bonnet any suggestions on this?

jumanji/training/networks/graph_coloring/actor_critic.py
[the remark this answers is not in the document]
clement-bonnet: I would recommend something along these lines
```python
        batch_size, num_nodes = observation.colors.shape
        colors_used = jnp.isin(observation.colors, jnp.arange(num_nodes))
        color_embeddings = hk.Linear(self.model_size)(
            colors_used[..., None].astype(float)
        )  # Shape (batch_size, num_colors, 128)

        nodes_colored = observation.colors >= 0
        node_embeddings = hk.Linear(self.model_size)(
            nodes_colored[..., None].astype(float)
        )  # Shape (batch_size, num_colors, 128)
```

docs/env_anim/graph_coloring.gif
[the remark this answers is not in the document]
ElshadaiK: the colors are randomly assigned to edges by the way, not node-specific. But when the edges intersect, it's much harder to differentiate them

jumanji/environments/logic/graph_coloring/viewer.py
clement-bonnet: Agree! Having `pos` as an argument to both functions looks the best to me.

jumanji/training/networks/graph_coloring/actor_critic.py
clement-bonnet: Am I missing something or is this not exactly the transpose of one another?
I.e. `colors_cross_nodes_mask[i, j]` is `True` if node `j` has to be colored by color `i`. And `nodes_cross_colors_mask[i, j]` is `True` if node `i` has to be colored by color `j`.

Am I missing something?

jumanji/environments/logic/graph_coloring/env.py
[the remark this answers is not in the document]
ElshadaiK: The Viewer class is implemented using Fruchterman-Reingold force-directed algorithm to get the spring layout which is: 
```
The algorithm computes a layout by simulating the graph as a physical system,
        where nodes are repelling each other and edges are attracting connected nodes.
        The method minimizes the energy of the system over several iterations.
```
So the viewer would like to know the number of nodes available. Of course I can take the num_nodes from the State but that would be calculating the scale and color map every time render or animate are called. Do we prefer to do that as the viewer has now become a parameter passed to the env?

jumanji/training/networks/graph_coloring/actor_critic.py
ElshadaiK: Yes, you're right. Gotcha!

jumanji/environments/logic/graph_coloring/generator.py
```diff
@@ -0,0 +1,126 @@
...
+
+class RandomGenerator(Generator):
+    """A generator for random graphs in the context of GraphColoring problems,
+    based on the Erdős-Rényi model (G(n, p)).
+
+    The adjacency matrix is generated such that the graph is undirected and loop-less.
+    The graph is generated with a specified number of nodes and percentage of connectivity,
+    which is used as a proxy for the edge probability in the Erdős-Rényi model.
+
+    """
```
sash-a, line 52:
```suggestion
    """
```

jumanji/environments/logic/graph_coloring/generator.py
```diff
@@ -0,0 +1,126 @@
...
+            percent_connected: A float between 0 and 1 representing the percentage of connections
+                in the graph compared to a fully connected graph.
+                The default value of `percent_connected` is 0.8.
+        """
+
+        self.num_nodes = num_nodes or 100
+        self.percent_connected = percent_connected or 0.8
+        assert (
+            self.percent_connected < 1
+        ), f"percent_connected={self.percent_connected} exceeds edges of a fully-connected graph."
```
sash-a, line 74: Just to be safe
```suggestion
        assert (
            0 < self.percent_connected < 1
        ), f"percent_connected={self.percent_connected} must be between 0 and 1."
```

jumanji/environments/logic/graph_coloring/generator.py
```diff
@@ -0,0 +1,126 @@
...
+        )
+
+        # Threshold the probabilities to create a boolean adjacency matrix.
+        adj_matrix = p_matrix < self.percent_connected
+
+        # Make sure the graph is undirected (symmetric) and without self-loops.
+        adj_matrix = jnp.tril(adj_matrix, k=-1)  # Keep only the lower triangular part.
+        adj_matrix += (
+            adj_matrix.T
+        )  # Copy the lower triangular part to the upper triangular part.
```
sash-a, line 111:
```suggestion
        # Copy the lower triangular part to the upper triangular part.
        adj_matrix += adj_matrix.T
```

sash-a (commented):
Looks great! Very minor style comments, but otherwise LGTM :fire:

jumanji/environments/logic/graph_coloring/env.py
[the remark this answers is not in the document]
ElshadaiK: @clement-bonnet & @coyettev,

Any comment on this or should I resolve the issue?

jumanji/training/networks/graph_coloring/actor_critic.py
ElshadaiK: @clement-bonnet,

Any comment on this or should I resolve the issue?

jumanji/environments/logic/graph_coloring/env.py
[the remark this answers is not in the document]
coyettev: I would say that yes it could be taken from the state, as computing the scale and color map should be quick, but happy to see what @clement-bonnet thinks

docs/environments/graph_coloring.md
```diff
@@ -0,0 +1,56 @@
+# Graph Coloring Environment
+
+<p align="center">
+    <img src="../env_img/graph_coloring.png" width="500"/>
+</p>
+
+We provide here a Jax JIT-able implementation of the Graph Coloring environment.
+
+GraphColoring is a combinatorial optimization problem where the objective is to assign a color to each vertex of a graph in such a way that no two adjacent vertices share the same color. The problem is usually formulated as minimizing the number of colors used. The Graph Coloring environment is an episodic, single-agent setting that allows for the exploration of graph coloring algorithms and reinforcement learning methods.
```
dluo96, line 9:
```suggestion
Graph coloring is a combinatorial optimization problem where the objective is to assign a color to each vertex of a graph in such a way that no two adjacent vertices share the same color. The problem is usually formulated as minimizing the number of colors used. The `GraphColoring` environment is an episodic, single-agent setting that allows for the exploration of graph coloring algorithms and reinforcement learning methods.
```

docs/environments/graph_coloring.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Observation
+
+The observation in the GraphColoring environment includes information about the graph, the colors assigned to the vertices, the action mask, and the current node index.
+
+- `graph`: jax array (bool) of shape (num_nodes, num_nodes), representing the adjacency matrix of the graph.
```
dluo96, line 15:
```suggestion
- `graph`: jax array (bool) of shape `(num_nodes, num_nodes)`, representing the adjacency matrix of the graph.
```

docs/environments/graph_coloring.md
```diff
@@ -0,0 +1,56 @@
...
+  - For example, a random observation of the graph adjacency matrix:
+
+        ```[[False,  True, False,  True],
+        [ True, False,  True, False],
+        [False,  True, False,  True],
+        [ True, False,  True, False]]```
+
+- `colors`: a JAX array (int32) of shape (num_nodes,), representing the current color assignments for the vertices. Initially, all elements are set to -1, indicating that no colors have been assigned yet.
```
dluo96, line 23:
```suggestion
- `colors`: a JAX array (int32) of shape `(num_nodes,)`, representing the current color assignments for the vertices. Initially, all elements are set to -1, indicating that no colors have been assigned yet.
```

docs/environments/graph_coloring.md
```diff
@@ -0,0 +1,56 @@
...
+  - For example, an initial color assignment:
+    ```[-1, -1, -1, -1]```
+
+- `action_mask`: a JAX array of boolean values, shaped (num_colors,), which indicates the valid actions in the current state of the environment. Each position in the array corresponds to a color. True at a position signifies that the corresponding color can be used to color a node, while False indicates the opposite.
```
dluo96, line 27: Could we add ticks to all the shapes?
```suggestion
- `action_mask`: a JAX array of boolean values, shaped `(num_colors,)`, which indicates the valid actions in the current state of the environment. Each position in the array corresponds to a color. True at a position signifies that the corresponding color can be used to color a node, while False indicates the opposite.
```

docs/environments/graph_coloring.md
```diff
@@ -0,0 +1,56 @@
...
+  - For example, for 4 number of colors available:
+    ```[True, False, True, False]```
+
+- `current_node_index`: an integer representing the current node being colored.
+  - For example, an initial current_node_index might be 0.
+
+## Action
+
+The action space is a DiscreteArray of integer values in [0, 1, ..., num_colors - 1]. Each action corresponds to assigning a color to the current node.
```
dluo96, line 36:
```suggestion
The action space is a DiscreteArray of integer values in `[0, 1, ..., num_colors - 1]`. Each action corresponds to assigning a color to the current node.
```

docs/environments/graph_coloring.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Reward
+
+The reward in the GraphColoring environment is given as follows:
```
dluo96, line 40:
```suggestion
The reward in the `GraphColoring` environment is given as follows:
```

docs/environments/graph_coloring.md
```diff
@@ -0,0 +1,56 @@
...
+
+- `sparse reward`: a reward is provided at the end of the episode and equals the negative of the number of unique colors used to color all vertices in the graph.
+
+The agent's goal is to find a valid coloring using as few colors as possible while avoiding conflicts with adjacent nodes.
+
+## Episode Termination
+
+The goal of the agent is to find a valid coloring using as few colors as possible. An episode in the graph coloring environment can terminate under two conditions:
+
+1. All nodes have been assigned a color: the environment iteratively assigns colors to nodes. When all nodes have a color assigned (i.e., there are no nodes with a color value of -1), the episode ends. This is the natural termination condition and ideally the one we'd like the agent to achieve. In the code, this is checked by the line all_nodes_colored = jnp.all(colors >= 0).
```
dluo96, line 50: Do we need to add the detail of how it's implemented in the code? Maybe we can remove it? I don't we have provided this detail for other environments

dluo96:
```suggestion
1. All nodes have been assigned a color: the environment iteratively assigns colors to nodes. When all nodes have a color assigned (i.e., there are no nodes with a color value of -1), the episode ends. This is the natural termination condition and ideally the one we'd like the agent to achieve.
```

jumanji/environments/logic/graph_coloring/env.py
```diff
@@ -0,0 +1,310 @@
...
+    def __init__(
+        self,
+        generator: Optional[Generator] = None,
+        viewer: Optional[Viewer[State]] = None,
+    ):
+        """Instantiate a `GraphColoring` environment.
+
+        Args:
+            generator: callable to instantiate environment instances.
+                Defaults to `RandomGenerator`.
```
dluo96, line 94: Is it worth specifying which parameters this random generator is instantiated with? I.e. 100 nodes where 80% of the nodes are connected? I think other environments have done that, but do correct me if I'm wrong

jumanji/environments/logic/graph_coloring/env.py
```diff
@@ -0,0 +1,310 @@
...
+        )
+        num_nodes = self.generator.num_nodes
+
+        # Create viewer used for rendering
+        self._env_viewer = viewer or GraphColoringViewer(
+            num_nodes=num_nodes, name="GraphColoring"
+        )
+
+    def __repr__(self) -> str:
+        """Returns: str: the string representation of the environment."""
```
dluo96, line 111: Maybe we don't need a doc string for this? I don't recall us doing this in other envs.

jumanji/environments/logic/graph_coloring/env.py
```diff
@@ -0,0 +1,310 @@
...
+        return f"GraphColouring({self.generator!r})"
```
dluo96, line 112: Why is the "!r" there?

dluo96, line 112: Why is the "!r" there?

jumanji/environments/logic/graph_coloring/env.py
```diff
@@ -0,0 +1,310 @@
...
+
+    def reset(self, key: chex.PRNGKey) -> Tuple[State, TimeStep[Observation]]:
+        """Resets the environment to an initial state.
+
+        Returns:
+            the initial state and timestep.
```
dluo96, line 118:
```suggestion
            The initial state and timestep.
```

jumanji/environments/logic/graph_coloring/env.py
```diff
@@ -0,0 +1,310 @@
...
+
+        Returns:
+            state: the new state of the environment.
+            timestep: the next timestep.
+        """
+        num_nodes = self.generator.num_nodes
+        # Get the valid actions for the current state.
+        valid_actions = state.action_mask
+        # Check if the chosen action is invalid (not in valid_actions).
+        invalid_action_taken = jnp.logical_not(valid_actions[action])
```
dluo96, line 160: NIT:
```suggestion
        num_nodes = self.generator.num_nodes
        
        # Get the valid actions for the current state.
        valid_actions = state.action_mask
        
        # Check if the chosen action is invalid (not in valid_actions).
        invalid_action_taken = jnp.logical_not(valid_actions[action])
```

jumanji/environments/logic/graph_coloring/env.py
```diff
@@ -0,0 +1,310 @@
...
+            - adj_matrix: BoundedArray (bool) of shape (num_nodes, num_nodes).
+                Represents the adjacency matrix of the graph.
+            - action_mask: BoundedArray (bool) of shape (num_nodes,).
+                Represents the valid actions in the current state.
+            - colors: BoundedArray (int32) of shape (num_nodes,).
+                Represents the colors assigned to each node.
+            - current_node_index: BoundedArray (int32) of shape ().
+                Represents the index of the current node.
+        """
+        num_nodes = self.generator.num_nodes
```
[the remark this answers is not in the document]
dluo96:
```suggestion
```

dluo96 (changes requested):
Thank you for your contribution @ElshadaiK! Have left a partial review of this PR.

jumanji/environments/logic/graph_coloring/env.py
```diff
@@ -0,0 +1,310 @@
...
+
+        # Update the colors array with the chosen action.
+        colors = state.colors.at[state.current_node_index].set(action)
+
+        # Determine if all nodes have been assigned a color
+        all_nodes_colored = jnp.all(colors >= 0)
+
+        # Calculate the reward
+        unique_colors_used = jnp.unique(colors, size=num_nodes, fill_value=-1)
+        num_unique_colors = jnp.count_nonzero(unique_colors_used >= 0)
```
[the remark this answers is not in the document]
ElshadaiK: yes it is JAX jittable since the size is statically specified like you said

jumanji/environments/logic/graph_coloring/env.py
```diff
@@ -0,0 +1,310 @@
...
+        next_node_index = (state.current_node_index + 1) % num_nodes
+
+        next_state = State(
+            adj_matrix=state.adj_matrix,
+            colors=colors,
+            current_node_index=next_node_index,
+            action_mask=valid_actions,
+            key=state.key,
+        )
+        next_state.action_mask = self._get_valid_actions(next_state)
```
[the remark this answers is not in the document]
ElshadaiK: This is because to even get the new action mask, we need the new state which contains the next node index.

jumanji/environments/logic/graph_coloring/env.py
ElshadaiK: The Cleaner and Connector environments specified the default values for example :)

jumanji/environments/logic/graph_coloring/env.py
ElshadaiK: This is on the assumption that we can also use another generator and if that's the case, we need to make sure we call the `__repr__` method. (Else we'll be calling `str(self.generator)` which could be different.

dluo96: I'm wondering if simply using `repr(self.generator)` might be a bit clearer here - what do you think?

ElshadaiK: yeah that makes sense tbh

jumanji/environments/logic/graph_coloring/env.py
ElshadaiK: we have that right on the 95th line though 😄

dluo96: Ah my bad, I'd totally missed that! All good, will resolve this comment 😄

jumanji/training/networks/graph_coloring/actor_critic.py
```diff
@@ -0,0 +1,293 @@
...
+        colors_used = jnp.isin(observation.colors, jnp.arange(num_nodes))
+        color_embeddings = hk.Linear(self.model_size)(
+            colors_used[..., None].astype(float)
+        )  # Shape (batch_size, num_colors, 128)
+
+        nodes_colored = observation.colors >= 0
+
+        color_embeddings = self.embed_colors(
+            colors_used
+        )  # Shape (batch_size, num_colors, 128)
```
clement-bonnet, line 131: This does not seem needed since already defined above. I believe we could remove these `embed_colors` and `embed_nodes` methods that only implements one operation anyway.

jumanji/training/networks/graph_coloring/actor_critic.py
```diff
@@ -0,0 +1,293 @@
...
+        )  # Shape (batch_size, transformer_num_heads, num_nodes, num_nodes)
+
+        colors_array = jnp.expand_dims(
+            observation.colors, axis=1
+        )  # Shape (batch_size, 1, num_nodes)
+        color_indices = jnp.arange(observation.colors.shape[1])  # Shape (num_colors,)
+        color_indices = jnp.expand_dims(color_indices, axis=0)  # Shape (1, num_colors)
+        color_indices = jnp.expand_dims(
+            color_indices, axis=-1
+        )  # Shape (1, num_colors, 1)
```
clement-bonnet, line 147:
```suggestion
        color_indices = color_indices[None, :, None]  # Shape (1, num_colors, 1)
```

jumanji/training/networks/graph_coloring/actor_critic.py
```diff
@@ -0,0 +1,293 @@
...
+
+        colors_cross_nodes_mask = (
+            colors_array == color_indices
+        )  # Shape (batch_size, num_colors, num_nodes)
+
+        # Expand and repeat along the transformer_num_heads axis
```
clement-bonnet, line 153:
```suggestion
        # Expand along the transformer_num_heads axis
```

jumanji/training/networks/graph_coloring/actor_critic.py
```diff
@@ -0,0 +1,293 @@
...
+        )  # Shape (batch_size, 1, num_colors, num_nodes)
+
+        colors_array = jnp.expand_dims(
+            observation.colors, axis=-1
+        )  # Shape (batch_size, num_nodes, 1)
+        color_indices = jnp.arange(observation.colors.shape[1])  # Shape (num_colors,)
+        color_indices = jnp.expand_dims(color_indices, axis=0)  # Shape (1, num_colors)
+        color_indices = jnp.expand_dims(
+            color_indices, axis=0
+        )  # Shape (1, 1, num_colors)
```
clement-bonnet, line 165:
```suggestion
        color_indices = color_indices[None, None]  # Shape (1, 1, num_colors)
```

jumanji/training/networks/graph_coloring/actor_critic.py
```diff
@@ -0,0 +1,293 @@
...
+
+        nodes_cross_colors_mask = (
+            colors_array == color_indices
+        )  # Shape (batch_size, num_nodes, num_colors)
+
+        # Expand and repeat along the transformer_num_heads axis
```
clement-bonnet, line 171:
```suggestion
        # Expand along the transformer_num_heads axis
```

jumanji/training/networks/graph_coloring/actor_critic.py
```diff
@@ -0,0 +1,293 @@
...
+        color_embeddings = self.embed_colors(
+            colors_used
+        )  # Shape (batch_size, num_colors, 128)
+        node_embeddings = self.embed_nodes(
+            nodes_colored
+        )  # Shape (batch_size, num_colors, 128)
+
+        mask = self._make_self_attention_mask(
+            observation.adj_matrix
+        )  # Shape (batch_size, transformer_num_heads, num_nodes, num_nodes)
```
clement-bonnet, line 138:
```suggestion
        )  # Shape (batch_size, 1, num_nodes, num_nodes)
```

clement-bonnet (commented):
LGTM! Just need to validate the training curve with you.

clement-bonnet (commented):
LGTM! Just need to validate the training curve with you.

jumanji/environments/logic/graph_coloring/env.py
```diff
@@ -0,0 +1,313 @@
...
+        """Specification of the action for the `GraphColoring` environment.
+
+        Returns:
+            action_spec: specs.DiscreteArray object
+        """
+        return specs.DiscreteArray(
+            num_values=self.num_nodes, name="action", dtype=jnp.int32
+        )
+
+    def _get_valid_actions(self, state: State) -> chex.Array:
```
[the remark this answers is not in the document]
ElshadaiK: Noted!

jumanji/environments/logic/graph_coloring/env.py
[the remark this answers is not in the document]
clement-bonnet: Sorry for coming late to this. If it's only for the feature map, I think it could be inferred at render time instead of at init time. So agree with @coyettev, it would be better to not have `num_nodes` a property of the viewer.

docs/environments/graph_coloring.md
```diff
@@ -0,0 +1,56 @@
...
+
+The goal of the agent is to find a valid coloring using as few colors as possible. An episode in the graph coloring environment can terminate under two conditions:
+
+1. All nodes have been assigned a color: the environment iteratively assigns colors to nodes. When all nodes have a color assigned (i.e., there are no nodes with a color value of -1), the episode ends. This is the natural termination condition and ideally the one we'd like the agent to achieve.
+
+2. Invalid action is taken: an action is considered invalid if it tries to assign a color to a node that is not within the allowed color set for that node at that time. The allowed color set for each node is updated after every action. If an invalid action is attempted, the episode immediately terminates and the agent receives a large negative reward. This encourages the agent to learn valid actions and discourages it from making invalid actions.
+
+## Registered Versions 📖
+
+- `GraphColoring-v0`: The default settings for the `GraphColoring` problem with a configurable number of nodes and connectivity.
```
clement-bonnet, line 56: We would need to refer to (i.e. provide the value of) the default settings

jumanji/environments/logic/graph_coloring/env.py
```diff
@@ -0,0 +1,318 @@
...
+
+        Args:
+            generator: callable to instantiate environment instances.
+                Defaults to `RandomGenerator` which generates graphs with
+                100 `num_nodes` and `percent_connected` equal to 0.8.
+            viewer: environment viewer for rendering.
+                Defaults to `GraphColoringViewer`.
+        """
+        self.generator = generator or RandomGenerator(
+            num_nodes=100, percent_connected=0.8
```
clement-bonnet, line 100: ?
```suggestion
            num_nodes=100, prop_connected=0.8
```

jumanji/environments/logic/graph_coloring/env.py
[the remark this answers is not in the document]
dluo96: I see 👍

jumanji/environments/logic/graph_coloring/generator.py
```diff
@@ -0,0 +1,105 @@
...
+
+import abc
+
+import chex
+import jax
+from jax import numpy as jnp
+
+
+class Generator(abc.ABC):
+    @property
```
dluo96, line 23: What was the reasoning for making this a property?

jumanji/environments/logic/graph_coloring/generator.py
```diff
@@ -0,0 +1,105 @@
...
+    @abc.abstractmethod
+    def num_nodes(self) -> int:
+        """Number of nodes of the problem instances generated.
+        Returns:
```
dluo96, line 27:
```suggestion
        """Number of nodes of the problem instances generated.
        
        Returns:
```

jumanji/environments/logic/graph_coloring/generator.py
```diff
@@ -0,0 +1,105 @@
...
+            `num_nodes` of the generated instances.
+        """
+
+    @abc.abstractmethod
+    def __call__(self, key: chex.PRNGKey) -> chex.Array:
+        """Generate a problem instance.
+
+        Args:
+            key: random key.
```
dluo96, line 36: Can we describe `key` as in other environments (referring to PRNG)?

jumanji/environments/logic/graph_coloring/generator.py
```diff
@@ -0,0 +1,105 @@
...
+
+        Returns:
+            An `adj_matrix` representing a problem instance.
+        """
+
+
+class RandomGenerator(Generator):
+    """A generator for random graphs in the context of GraphColoring problems,
```
dluo96, line 44:
```suggestion
    """A generator for random graphs in the context of graph coloring problems,
```

jumanji/environments/logic/graph_coloring/generator.py
```diff
@@ -0,0 +1,105 @@
...
+    based on the Erdős-Rényi model (G(n, p)).
+
+    The adjacency matrix is generated such that the graph is undirected and loop-less.
+    The graph is generated with a specified number of nodes and percentage of connectivity,
+    which is used as a proxy for the edge probability in the Erdős-Rényi model.
+    """
+
+    def __init__(self, num_nodes: int, edge_probability: float):
+        """
+        Initialize the RandomGraphColoringGenerator.
```
dluo96, line 54: Can you double check that doc strings are done consistently?
```suggestion
        """Initialize the RandomGraphColoringGenerator.
```

jumanji/environments/logic/graph_coloring/generator.py
```diff
@@ -0,0 +1,105 @@
...
+            f"percent connected={self.edge_probability * 100}%)"
+        )
+
+    def __call__(self, key: chex.PRNGKey) -> chex.Array:
+        """Generate a random graph adjacency matrix representing
+        the edges of an undirected graph using the Erdős-Rényi model G(n, p).
+        Args:
+            key: A Chex PRNGKey used for random number generation.
+        Returns:
+            adj_matrix: A boolean array of shape (num_nodes, num_nodes) representing
```
dluo96, line 86:
```suggestion
        the edges of an undirected graph using the Erdős-Rényi model G(n, p).
        
        Args:
            key: PRNGKey used for stochasticity in the generation process.
            
        Returns:
            adj_matrix: a boolean array of shape (num_nodes, num_nodes) representing
```

jumanji/environments/logic/graph_coloring/generator.py
```diff
@@ -0,0 +1,105 @@
...
+        p_matrix = jax.random.uniform(
+            key=edge_key, shape=(self.num_nodes, self.num_nodes)
+        )
+
+        # Threshold the probabilities to create a boolean adjacency matrix.
+        adj_matrix = p_matrix < self.edge_probability
+
+        # Make sure the graph is undirected (symmetric) and without self-loops.
+        adj_matrix = jnp.tril(adj_matrix, k=-1)  # Keep only the lower triangular part.
+        # Copy the lower triangular part to the upper triangular part.
```
dluo96, line 102:
```suggestion
        adj_matrix = jnp.tril(adj_matrix, k=-1)  # Keep only the lower triangular part.
         
        # Copy the lower triangular part to the upper triangular part.
```

jumanji/training/networks/graph_coloring/actor_critic.py
```diff
@@ -0,0 +1,278 @@
...
+        self.model_size = transformer_num_heads * transformer_key_size
+
+    def _make_self_attention_mask(self, adj_matrix: chex.Array) -> chex.Array:
+        # Expand on the head dimension.
+        mask = jnp.expand_dims(adj_matrix, axis=-3)
+        return mask
+
+    def __call__(self, observation: Observation) -> chex.Array:
+        """
+        Transforms the observation using a series of transformations.
```
dluo96, line 89:
```suggestion
        """Transforms the observation using a series of transformations.
```

jumanji/training/networks/graph_coloring/actor_critic.py
```diff
@@ -0,0 +1,278 @@
...
+        It further creates two masks to track the relation between colors and nodes.
+        Then the function applies a series of transformer blocks to compute:
+            self-attention on nodes and colors and the cross-attention between nodes and colors.
+        Finally, it extracts the embedding for the current node and computes a new embedding.
+
+        Args:
+            observation: The observation to be transformed.
+
+        Returns:
+            new_embedding: The transformed observation.
```
dluo96, line 108:
```suggestion
            observation: the observation to be transformed.

        Returns:
            new_embedding: the transformed observation.
```

dluo96 (changes requested):
Left a few final comments, otherwise LGTM! 🚀

dluo96 (changes requested):
Left a few final comments, otherwise LGTM! 🚀

dluo96 (changes requested):
Left a few final comments, otherwise LGTM! 🚀

jumanji/environments/logic/graph_coloring/generator.py
ElshadaiK: we need to get num_nodes for several operations in the environment, and it was suggested that I make it a generator property rather than generator spec.

clement-bonnet (dismissed):
Looks good to me, thank you!

jumanji/environments/logic/graph_coloring/viewer_test.py
```diff
@@ -0,0 +1,70 @@
...
+        action = jnp.array(0)
+        new_state, _ = graph_coloring.step(state, action)
+        states.append(new_state)
+        state = new_state
+
+    save_path = "animation_test.gif"
+    graph_coloring.animate(states, interval=500, save_path=save_path)
+
+    assert os.path.exists(save_path)
+    os.remove(save_path)
```
dluo96, line 70: Could we use the built-in pytest fixture `tmp_path` in this test? That would automatically be ensure clean-up and remove the need for `os.remove`.

dluo96 (approved):
LGTM! 🚀

dluo96 (dismissed):
LGTM! 🚀

jumanji/environments/logic/graph_coloring/viewer_test.py
ElshadaiK: Done!

clement-bonnet (approved):

clement-bonnet (approved):

dluo96 (approved):

PR #139, opened 2023-05-18
feat(knapsack): generator

surana01:
Implement a `Generator` class for the Knapsack environment which is responsible for generating the Knapsack instances. This will include the following additions:
- A `generator.py` script added to the `knapsack` folder which implements an abstract class `Generator` and a `RandomGenerator` to create a random uniform Knapsack instance. 
- A `generator_test.py` script to test the generator. 
- Modifications to `env.py` to include the generator in the `init` method and changes to `conftest.py` to include a `DummyGenerator` for testing purposes.

### DIFF
--- a/jumanji/environments/packing/knapsack/generator.py
+++ b/jumanji/environments/packing/knapsack/generator.py
@@ -1,0 +1,86 @@
+# Copyright 2022 InstaDeep Ltd. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


jumanji/environments/packing/knapsack/generator.py
```diff
@@ -0,0 +1,86 @@
...
+
+        Args:
+            key: jax random key in case stochasticity is used in the instance generation process.
+
+        Returns:
+            A `Knapsack` environment state.
+        """
+
+
+class RandomGenerator(Generator):
```
clement-bonnet, line 55: Should we call it `UniformGenerator` as well?

surana01: I was hesitant here because for TSP and CVRP it is common to have random uniform (RUE) instances, but I have not seen the same for KP, so I am not sure here whether we should call it `UniformGenerator`. But happy to change it if you think it makes sense here.

clement-bonnet (approved):
LGTM :rocket:

PR #142, opened 2023-05-21
fix(parametric_action_distribution): sum kl divergence over event_ndims in parametric action distribution

clement-bonnet:


jumanji/training/networks/parametric_distribution.py
```diff
@@ -110,9 +110,16 @@ def kl_divergence(
         self, parameters: chex.Array, other_parameters: chex.Array
     ) -> chex.Array:
         """KL divergence is invariant with respect to transformation by the same bijector."""
+        if not isinstance(self._postprocessor, IdentityBijector):
+            raise ValueError("KL is not implemented for non-identity bijectors.")
```
djbyrne, line 114: I think this is kind of unclear as a user message. Maybe replace with 

"The current post_processor used is a non-identity bijector which does not implement KL"

jumanji/training/networks/parametric_distribution.py
```diff
@@ -110,9 +110,16 @@ def kl_divergence(
...
         dist = self.create_dist(parameters)
         other_dist = self.create_dist(other_parameters)
-        return dist.kl_divergence(other_dist)
+        kl_divergence = dist.kl_divergence(other_dist)
+        if self._event_ndims == 1:
+            kl_divergence = jnp.sum(kl_divergence, axis=-1)
+        else:
+            assert self._event_ndims == 0
```
djbyrne, line 121: can we replace this assert with a raise value error?

jumanji/training/networks/parametric_distribution.py
clement-bonnet: Done

djbyrne (approved):

PR #145, opened 2023-05-25
fix(training): connector num_agents in networks

clement-bonnet:
The `Connector` networks are confusing `num_agents` with `num_actions`. This is not a problem in the registered version since both are 5. However, it becomes a problem when one trains on a different number of agents.

I also fixed a `squeeze` in the evaluator file.

### DIFF
--- a/jumanji/training/networks/connector/actor_critic.py
+++ b/jumanji/training/networks/connector/actor_critic.py
@@ -50,6 +50,7 @@ def make_actor_critic_networks_connector(
     parametric_action_distribution = MultiCategoricalParametricDistribution(
         num_values=num_values
     )
+    # num_values is of shape (num_agents,) and contains num_actions everywhere.
     num_agents = num_values.shape[0]
     num_actions = num_values[0]
     policy_network = make_actor_network_connector(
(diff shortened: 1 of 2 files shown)


jumanji/training/networks/connector/actor_critic.py
```diff
@@ -50,8 +50,11 @@ def make_actor_critic_networks_connector(
     parametric_action_distribution = MultiCategoricalParametricDistribution(
         num_values=num_values
     )
+    num_agents = num_values.shape[0]
+    num_actions = num_values[0]
```
djbyrne, line 54: NIT: maybe add an inline comment on what the shape, contents of num_values is so it is more intuitive

djbyrne (approved):

sash-a (dismissed):
Nice catch :rocket:

jumanji/training/networks/connector/actor_critic.py
```diff
@@ -50,8 +50,11 @@ def make_actor_critic_networks_connector(
     parametric_action_distribution = MultiCategoricalParametricDistribution(
         num_values=num_values
     )
```
clement-bonnet, line 52:
```suggestion
    )
    # num_values is an array of shape (num_agents,) and contains num_actions everywhere.
```

jumanji/training/networks/connector/actor_critic.py
clement-bonnet: Done :heavy_check_mark:

djbyrne (approved):

PR #154, opened 2023-06-02
docs: image fixes

clement-bonnet:


docs/environments/job_shop.md
```diff
@@ -1,7 +1,7 @@
 # JobShop Environment
 
 <p align="center">
-        <img src="../env_anim/job_shop_annotated.gif" width="600"/>
+        <img src="../img/job_shop_annotated.gif" width="600"/>
```
dluo96, line 4: Why the switch?

dluo96 (approved):
LGTM!

docs/environments/job_shop.md
clement-bonnet: We want to keep in `env_anim` one GIF per environment. JobShop already has such a gif called `job_shop.gif`. Hence we either remove one or move this one to the more general `img` folder that does not have restrictions.

dluo96: Agreed!

PR #158, opened 2023-06-04
docs: update online doc with new environments

### LINKED ISSUE #148: build: remove `jaxlib`
You do not need to explicitly state the version of jaxlib because `pip install jax[<device>]` will automatically install the correct `jaxlib`. This is the culprit line:
https://github.com/instadeepai/jumanji/blob/b16cf5dcde88a73b0c8f56f93b97813105cb99ec/requirements/requirements.txt#L5

### LINKED ISSUE #155: docs: update online doc with new environments
The newly added environments do not appear on the `mk_docs.yaml`, which means they are not rendered on the online documentation.

We must update the `mk_docs.yaml` API reference and environments.

clement-bonnet:
fixes #155 

- add new environments to the online doc
- update `RobotWarehouse` and `Tetris` PNGs and GIFs
- display all 18 GIFs in the readme


README.md
```diff
@@ -19,22 +19,24 @@
...
+<div align="center">
+  <img src="docs/env_anim/connector.gif" alt="Connector" width="30%" style="margin: 5px;">
+  <img src="docs/env_anim/snake.gif" alt="Snake" width="30%" style="margin: 5px;">
+  <img src="docs/env_anim/cleaner.gif" alt="Cleaner" width="30%" style="margin: 5px;">
+</div>
+
+<div align="center">
+  <img src="docs/env_anim/job_shop.gif" alt="JobShop" width="30%" style="margin: 5px;">
+  <img src="docs/env_anim/bin_pack.gif" alt="BinPack" width="30%" style="margin: 5px;">
+  <img src="docs/env_anim/tetris.gif" alt="Tetris" width="18%" style="margin: 5px 6.5%;">
```
djbyrne, line 31: Should the dimensions of tetris not be uniform with the rest? the middle row of the grid looks off, maybe this is why?

djbyrne (changes requested):

README.md
TristanKalloniatis, line 31: Any reason for the different width/style here compared to all the others?

README.md
```diff
@@ -19,22 +19,24 @@
...
+</div>
+
+<div align="center">
+  <img src="docs/env_anim/rubiks_cube.gif" alt="RubiksCube" width="30%" style="margin: 5px;">
```
TristanKalloniatis, line 35: Dunno your thoughts on alt text, the env is called RubiksCube but the puzzle is called Rubik's Cube

README.md
clement-bonnet: To your first question, a similar comment was made above. To your second, this was to remain consistent with what has been before. We will make another update later to include all 18 GIFs in a display like on the paper. I think all environments should be on the page, otherwise it looks like there are only 9 environments implemented.

README.md
clement-bonnet: We are talking about environment names here, mappings class names to GIFs

README.md
TristanKalloniatis: OK but until then is there any reason not to list eg BinPack, JobShop, Tetris?

TristanKalloniatis: IMO this is an argument to not include Tetris in this iteration

djbyrne: I can't tell if Tristan is joking or not :sweat_smile:

djbyrne (dismissed):

README.md
clement-bonnet: Fair point Tristan.

clement-bonnet: Updated `Tetris` and `RobotWarehouse` GIFs to be square. I also added all the other environments to have all 18 GIFs.

clement-bonnet:
Here is the current look of the readme
Screenshot from 2023-06-05 18-58-44

TristanKalloniatis:
> Here is the current look of the readme Screenshot from 2023-06-05 18-58-44

Looks good, a couple quick questions:

1. Some of the renderings look pretty similar, especially cleaner and maze. So in this view looks slightly odd IMO.
2. I don't get the ordering in this collage

TristanKalloniatis:
Is it purely alphabetical, ie not separated by env "type"?

clement-bonnet:
> > Here is the current look of the readme Screenshot from 2023-06-05 18-58-44
> 
> Looks good, a couple quick questions:
> 
> 1. Some of the renderings look pretty similar, especially cleaner and maze. So in this view looks slightly odd IMO.
> 2. I don't get the ordering in this collage

1. Maze and Cleaner are under the hood using the same viewer abstraction, therefore they look similar indeed. I think it is even a good thing since you can show that things are easy to share. Either way, it is better to have all 18 environments that a selection that could make people think that only a few numbers of environments are implemented.
2. Ordering is by alphabetic order (to mimic the paper).

PR #164, opened 2023-06-06
feat(training): multi-worker training

clement-bonnet:
- adapt training code to run on multiple workers (e.g. TPU v4-16 and more)
- [x] tested on CPU
- [x] tested on TPU v4-8
- [x] tested on TPU v4-16 (2 workers)

### DIFF
--- a/jumanji/training/setup_train.py
+++ b/jumanji/training/setup_train.py
@@ -50,6 +50,7 @@ from jumanji.training.evaluator import Evaluator
 from jumanji.training.loggers import (
     Logger,
     NeptuneLogger,
+    NoOpLogger,
     TensorboardLogger,
     TerminalLogger,
 )
@@ -61,6 +62,9 @@ from jumanji.wrappers import VmapAutoResetWrapper
 
 def setup_logger(cfg: DictConfig) -> Logger:
     logger: Logger
+    # Log only once if there are multiple hosts on the pod.
+    if jax.process_index() != 0:
+        return NoOpLogger()
     if cfg.logger.type == "tensorboard":
         logger = TensorboardLogger(
             name=cfg.logger.name, save_checkpoint=cfg.logger.save_checkpoint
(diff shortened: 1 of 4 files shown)


clement-bonnet:
CPU
image

TPU v4-8
image

TPU v4-16
image

jumanji/training/setup_train.py
```diff
@@ -61,6 +62,9 @@
 
 def setup_logger(cfg: DictConfig) -> Logger:
     logger: Logger
+    # Log only once if there are multiple hosts on the pod.
```
dluo96, line 65: To be clear, _host_ and _worker_ are referring to the same thing, right?

jumanji/training/setup_train.py
```diff
@@ -393,20 +397,33 @@ def setup_training_state(
...
-    num_devices = jax.local_device_count()
+    num_local_devices = jax.local_device_count()
+    num_global_devices = jax.device_count()
+    num_workers = num_global_devices // num_local_devices
     reset_keys = jax.random.split(reset_key, agent.total_batch_size).reshape(
-        (num_devices, agent.batch_size_per_device, -1)
+        (
+            num_workers,
+            num_local_devices,
+            agent.total_batch_size // num_global_devices,
```
dluo96, line 407: Suggestion: could set this to a variable `batch_size_per_global_device = agent.total_batch_size // num_global_devices`?

jumanji/training/setup_train.py
```diff
@@ -393,20 +397,33 @@ def setup_training_state(
...
+        reset_keys_per_worker
     )
-    env_state, timestep = jax.pmap(env.reset, axis_name="devices")(reset_keys)
 
     # Initialize acting states.
-    acting_key_per_device = jax.random.split(acting_key, num_devices)
+    acting_key_per_device = jax.random.split(acting_key, num_global_devices).reshape(
+        num_workers, num_local_devices, -1
+    )
+    acting_key_per_device_worker = acting_key_per_device[jax.process_index()]
```
dluo96, line 420: A bit confused by the naming, specifically "_device_worker" - "worker" comes higher in the hierarchy, so maybe it should be `_worker_device`?

jumanji/training/train.py
```diff
@@ -103,6 +105,7 @@ def epoch_fn(training_state: TrainingState) -> Tuple[TrainingState, Dict]:
             # Training
             with train_timer:
                 training_state, metrics = epoch_fn(training_state)
+                jax.block_until_ready((training_state, metrics))
```
dluo96, line 108: Remind me why we are adding this?

dluo96 (approved):
Left a few comments, otherwise LGTM

jumanji/training/setup_train.py
djbyrne: we use "local_batch_size" in sebulba for this  :+1:

jumanji/training/train.py
djbyrne: I imagine it is just to make sure there is no async_dispatch calls coming back and that the epoch has actually finished when we call it, otherwise we might get misleading metrics

jumanji/training/setup_train.py
clement-bonnet: I guess host is more general than worker? But yes.

jumanji/training/train.py
clement-bonnet: It's so the timing is correct, indeed.

jumanji/training/setup_train.py
clement-bonnet: Done

jumanji/training/setup_train.py
clement-bonnet: Done

jumanji/training/setup_train.py
dluo96: Ok, maybe we should stick with "worker" then? Since we use that in variable names

jumanji/training/train.py
dluo96: Makes sense 👍

dluo96 (approved):
LGTM! 🚀

dluo96 (approved):
LGTM! 🚀

jumanji/training/setup_train.py
clement-bonnet: The logger is different from multi-worker training, isn't it? I will stick to this to go fast since they are the same thing.

PR #166, opened 2023-06-08
feat: updated checkpoint notebook to use checkpoints from hf model hub

djbyrne:
Checkpoints for all benchmarks can be found in InstaDeeps Model Hub [page ](https://huggingface.co/InstaDeepAI). 

The naming convention is jumanji-[ENV-NAME]-[ENV-VERSION]-[MODEL]-[TYPE] 

The checkpoint file has the naming convention [ENV-NAME]-[ENV-VERSION]_training_state
Potentially it will be better to have a common `training_state` so that we can programatically pull down checkpoints?


clement-bonnet:
Closing this as a solution was found in #165

PR #172, opened 2023-06-15
feat(2048): environment performance improvements

aar65537:
This PR improves the performance of the Game2048 environment. The improvements include

- Minimizing logic inside of `jax.lax.cond` and `jax.lax.switch`
- Using `jax.vmap` over `jax.lax.scan` where possible
- A new move implementation
- A can_move implementation that validates an action without mutating the board

|      | no vmap | vmap 10<sup>3</sup> | vmap 10<sup>6</sup> |
|:-----|-:|-:|-:|
| cpu  | 64.36%  | 201.80%  | 392.29% |
| cuda | 900.12% | 1923.08% | 706.87% |

The above figure shows the total performance improvement measured as percent increase in steps/sec. For more detailed benchmarking, see [here](https://github.com/aar65537/jumanji-benchmarks).

### DIFF
--- a/jumanji/environments/logic/game_2048/env.py
+++ b/jumanji/environments/logic/game_2048/env.py
@@ -24,6 +24,7 @@ from jumanji import specs
 from jumanji.env import Environment
 from jumanji.environments.logic.game_2048.types import Board, Observation, State
 from jumanji.environments.logic.game_2048.utils import (
+    move,
     move_down,
     move_left,
     move_right,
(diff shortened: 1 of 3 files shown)


clement-bonnet:
Hi @aar65537, thanks a lot for your suggestions for speed improvement! We will look into them, check that the environment's behavior has not changed, and get back to you shortly.

clement-bonnet (approved):
Thank you very much for the code and speed improvements! I have left a few comments for which I don't feel very strongly about.

clement-bonnet (approved):

PR #176, opened 2023-06-19
docs: update readme citation

clement-bonnet:
Citation exported from [https://arxiv.org/abs/2306.09884](https://arxiv.org/abs/2306.09884)


dluo96 (approved):
LGTM

PR #181, opened 2023-06-26
fix: latest chex and jax

clement-bonnet:
With `chex>0.1.5`, `chex.Array` is not compatible with `isinstance` used in training.


dluo96 (approved):
LGTM!

PR #184, opened 2023-07-05
fix bugs in specs test

George-Ogden:
fixes failing test 
duplicate of #183 with signed commit


clement-bonnet (approved):
Thank you for the fix! I am unsure why this bug has persisted until today.

PR #191, opened 2023-07-19
fix(roboticWarehouse): goal conditional

### LINKED ISSUE #190: Fix: robot warehouse rewarding non-goal states
The logic for checking whether a shelf is at a goal state upon delivery before providing a reward has a bug (i.e. shelves are being rewarded at none goal states). The suggested fix is to decompose the conditional (as was originally done).

arnupretorius:
Fix the goal conditional for reward in robotic warehouse.

Closes #190

### DIFF
--- a/jumanji/environments/routing/robot_warehouse/env.py
+++ b/jumanji/environments/routing/robot_warehouse/env.py
@@ -491,11 +491,12 @@ class RobotWarehouse(Environment[State]):
             return key, reward, request_queue, shelves
 
         # check if shelf is at goal position and in request queue
-        cond = jnp.logical_not(jnp.equal(shelf_id, 0))
-        cond = jnp.logical_and(cond, jnp.isin(shelf_id, request_queue + 1))
+        shelf_at_goal = (~jnp.equal(shelf_id, 0)) & jnp.isin(
+            shelf_id, request_queue + 1
+        )
 
         key, reward, request_queue, shelves = jax.lax.cond(
-            cond,
+            shelf_at_goal,
             reward_and_update_request_queue_if_shelf_in_goal,
             lambda k, r, rq, g, _: (k, r, rq, g),
             key,
--- a/jumanji/environments/routing/robot_warehouse/env.py
+++ b/jumanji/environments/routing/robot_warehouse/env.py
@@ -454,7 +454,7 @@ class RobotWarehouse(Environment[State]):
         Returns:
             a random key, updated reward, request queue and shelves.
         """
-        x, y = goal
+        y, x = goal
         shelf_id = grid[_SHELVES, x, y]
 
         def reward_and_update_request_queue_if_shelf_in_goal(
--- a/jumanji/environments/routing/robot_warehouse/env.py
+++ b/jumanji/environments/routing/robot_warehouse/env.py
@@ -491,7 +491,8 @@ class RobotWarehouse(Environment[State]):
             return key, reward, request_queue, shelves
 
         # check if shelf is at goal position and in request queue
-        cond = (shelf_id != 0) & jnp.isin(shelf_id, request_queue + 1)
+        cond = jnp.logical_not(jnp.equal(shelf_id, 0))
+        cond = jnp.logical_and(cond, jnp.isin(shelf_id, request_queue + 1))
 
         key, reward, request_queue, shelves = jax.lax.cond(
             cond,
--- a/jumanji/environments/routing/robot_warehouse/env.py
+++ b/jumanji/environments/routing/robot_warehouse/env.py
@@ -251,12 +251,6 @@ class RobotWarehouse(Environment[State]):
         # check for invalid action -> turn into noops
         actions = utils.get_valid_actions(action, state.action_mask)
 
-        # check for agent collisions
-        collisions = jax.vmap(functools.partial(utils.is_collision, grid))(
-            agents, actions
-        )
-        collision = jnp.any(collisions)
-
         # update agents, shelves and grid
         def update_state_scan(
             carry_info: Tuple[chex.Array, chex.Array, chex.Array, int], action: int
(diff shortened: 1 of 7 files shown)


jumanji/environments/routing/robot_warehouse/env.py
```diff
@@ -491,7 +491,8 @@ def reward_and_update_request_queue_if_shelf_in_goal(
             return key, reward, request_queue, shelves
 
         # check if shelf is at goal position and in request queue
-        cond = (shelf_id != 0) & jnp.isin(shelf_id, request_queue + 1)
+        cond = jnp.logical_not(jnp.equal(shelf_id, 0))
+        cond = jnp.logical_and(cond, jnp.isin(shelf_id, request_queue + 1))
```
sash-a, line 495: Could we change this to:

`shelf_at_goal = (~jnp.equal(shelf_id, 0)) & jnp.isin(shelf_id, request_queue + 1)`

Makes the `cond` variable more explicit and uses `~` and `&` instead of the jax functions

jumanji/environments/routing/robot_warehouse/env.py
```diff
@@ -271,6 +265,12 @@ def update_state_scan(
             update_state_scan, (grid, agents, shelves, 0), actions
         )
 
+        # check for agent collisions
+        collisions = jax.vmap(functools.partial(utils.is_collision, grid))(
+            agents, self.agent_ids
+        )
```
sash-a, line 271: Up to you, but I prefer

```suggestion
        collisions = jax.vmap(utils.is_collision, (None, 0, 0))(
            grid, agents, self.agent_ids
        )
```

jumanji/environments/routing/robot_warehouse/utils.py
```diff
@@ -82,10 +82,11 @@ def get_valid_action(action_mask: chex.Array, action: chex.Array) -> chex.Array:
...
 
-def is_collision(grid: chex.Array, agent: Agent, action: int) -> chex.Array:
-    """Calculate whether an agent is about to collide with another
-    entity. If the agent instead collides with another agent, the
-    episode terminates (this behavior is specific to this JAX version).
+def is_collision(grid: chex.Array, agent: Agent, agent_id: int) -> chex.Array:
+    """Calculate whether an agent collides with another
+    agent. If there is a collision, theepisode terminates
+    (this behavior is specific to the JAX version of the
+    robotic warehouse (RWARE) environment).
```
sash-a, line 89:
```suggestion
    """Calculate whether an agent collides with another
    agent. If there is a collision, the episode terminates
    (this behavior is specific to the JAX version of the
    robotic warehouse (RWARE) environment).
```

sash-a (commented):
Looks good, just some grammar and some really minor changes

jumanji/environments/routing/robot_warehouse/env.py
arnupretorius: Thanks Sasha, when making these suggestions did you rerun the test? Since the issue fixed here crept in by accepting a similar suggestion, so I'm hesitant to do it again. Similar with the one suggested below. At this stage, since it is so minor, I'm inclined to keep what works. :)

clement-bonnet: I lean towards @sash-a 's, yet both expressions are probably converted to the same XLA statement, so happy to keep as is.

jumanji/environments/routing/robot_warehouse/env.py
clement-bonnet: I'm agnostic to this, up to you.

jumanji/environments/routing/robot_warehouse/env.py
```diff
@@ -454,7 +454,7 @@ def _update_reward_and_request_queue(
         Returns:
             a random key, updated reward, request queue and shelves.
         """
-        x, y = goal
+        y, x = goal
```
clement-bonnet, line 457: Was this an error as well? Why are coordinates inverted? In Jumanji, we usually follow the (x,y) convention instead of the video-game (y,x) one.

jumanji/environments/routing/robot_warehouse/env.py
clement-bonnet, line 495: Why is this code being changed? The original looks better.

clement-bonnet (approved):
Looks good to me. Would it be possible to start a few steps of a2c training (with default params) and seed either 1, 2 or 3 to make sure the behavior is fixed?

jumanji/environments/routing/robot_warehouse/env.py
arnupretorius: Reverted to Sasha's version

jumanji/environments/routing/robot_warehouse/env.py
arnupretorius: Keeping as is.

jumanji/environments/routing/robot_warehouse/env.py
arnupretorius: Yes, this was a bug. The reversal is a quirk from the original environment in terms of how they set up the initial grid. See here on how they all invert when enumerating goal positions: https://github.com/semitable/robotic-warehouse/blob/7b83fe04ec5b57523bc4d6b6c1c179c48849a3f7/rware/warehouse.py#L807

arnupretorius:
> Looks good to me. Would it be possible to start a few steps of a2c training (with default params) and seed either 1, 2 or 3 to make sure the behavior is fixed?

We ran some experiments to confirm, everything works as expected now. In essence, the problem remained the same but with inverted goal positions which caused a mismatch between the rendering. In the runs we did, the agents trained and the rendering looked good. I will also add an updated gif to this PR.

sash-a (approved):
LGTM :smile:

clement-bonnet (approved):
Thank you for the fixes! :rocket:

RuanJohn (approved):
Looks good! Thank you @arnupretorius :fire:

PR #194, opened 2023-08-01
fix(maze): fixed row and col in _compute_action_mask function.

danielpalen:
Row and column were mixed up. In mazes where num_rows != num_cols, this would lead to incorrect action masks. This was fixed and the naming was changed from (x, y) to (row, col) in order to improve readability.


PDuckworth (approved):
Looks good to me :+1:

clement-bonnet (approved):
Thank you for improving this function.

PR #196, opened 2023-08-17
docs: fix links at top of README

George-Ogden:
The links now work in the README.

PR #197, opened 2023-08-18
fix(CI): unpin mkdocs and plugin versions

sash-a:
Github workflows aren't passing because of an error in `mkdocs`, one of the plugins requires `mkdocs >= 1.4`, we're pinning at `1.2.3` I unpinned `mkdocs` and unpinned all plugins so we can stay up to date.

Tested locally and seems like the docs are working

### DIFF
--- a/requirements/requirements-dev.txt
+++ b/requirements/requirements-dev.txt
@@ -4,13 +4,13 @@ flake8==3.9.2
 importlib-metadata<5.0
 isort==5.11.5
 livereload
-mkdocs==1.2.3
+mkdocs
 mkdocs-git-revision-date-plugin
 mkdocs-include-markdown-plugin
 mkdocs-material
-mkdocs-mermaid2-plugin==0.6.0
-mkdocstrings==0.18.0
-mknotebooks==0.7.1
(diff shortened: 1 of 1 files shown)


requirements/requirements-dev.txt
```diff
@@ -4,13 +4,13 @@ flake8==3.9.2
 importlib-metadata<5.0
 isort==5.11.5
 livereload
-mkdocs==1.2.3
+mkdocs
```
djbyrne, line 7: It might be better to set a range for this? mkdocs >= 1.4.0

clement-bonnet (dismissed):
Awesome, thank you for the fix!

clement-bonnet (commented):
I think this got fixed in #196 [https://github.com/instadeepai/jumanji/commit/9e8aad624301311f47883621f6170cebfb62ec14](https://github.com/instadeepai/jumanji/commit/9e8aad624301311f47883621f6170cebfb62ec14)

clement-bonnet:
I will close this for now as it seems the issue got fixed in #196. Please, re-open if the issue arises again.

PR #204, opened 2023-09-07
fix: gh runners to use on-prem.

callumtilbury:
As per @alaterre's request. I think I've done everything correctly—will check when the PR opens, as I can't run self-hosted CI jobs from my personal acc 😅

### DIFF
--- a/.github/workflows/docs_deploy.yml
+++ b/.github/workflows/docs_deploy.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   docs:
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
     container:
       image: python:3.8.10
     steps:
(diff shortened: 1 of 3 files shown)


clement-bonnet:
Hey! Thanks for the PR. Just wondering, why are we doing that?

callumtilbury:
Hey @clement-bonnet! Alex has asked us to use self-hosted runners for all InstaDeep CI—perhaps he can clarify 👍🏻

callumtilbury:
> I wonder if this is the right thing to do here

It's a good question, @clement-bonnet. Still waiting to hear back from @alaterre.

In the interim, it's worth noting that the CI jobs aren't being picked up:
image
I may have made a syntax error somewhere, but the CI jobs _are_ running on other ID projects.

clement-bonnet:
@callumtilbury do we have an update on this? I'm closing this for now but please reopen if we have an update :)

callumtilbury:
thanks @clement-bonnet! Following up with @alaterre now :)

PR #206, opened 2023-09-08
fix: reward and discount spec not in wrapper

sash-a:
The `reward_spec` and `discount_spec` methods are not in `Wrapper`. As such when wrapping an environment with a reward/discount spec that is not the default (`shape=()`), the reward and discount specs are set to the default unless they are specified again in the wrapper.

For example
```python
class MyWrapper(Wrapper):
    def reset(self, key):
        do_some_stuff()
        return self._env.reset(key)

env = Connector()
wrapped = MyWrapper(env)

env.reward_spec() == wrapped.reward_spec()  # <-- false
```

### DIFF
--- a/jumanji/wrappers.py
+++ b/jumanji/wrappers.py
@@ -96,6 +96,14 @@ class Wrapper(Environment[State], Generic[State]):
         """Returns the action spec."""
         return self._env.action_spec()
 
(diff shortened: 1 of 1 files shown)


clement-bonnet (approved):
Well spotted, thanks!

PR #213, opened 2023-10-20
Update registration.md

navh:


sash-a:
I don't see an issue with the current approach. I can run `from jumanji import register`. So I'm closing the PR, feel free to re-open it if there's something I missed :smile:

PR #217, opened 2024-01-16
Fix: issue in dtype of grid in cleaner env

raphaelavalos:
Hello,

I found a bug in the cleaner environment. The dtype of the grid was incorrect and it was failing the tests. The alternative fix would be to change the spec of the grid to `int32`.

Best,
Raphael

### DIFF
--- a/docs/environments/cleaner.md
+++ b/docs/environments/cleaner.md
@@ -20,7 +20,7 @@ always start in the top left corner of the maze.
 ## Observation
 The **observation** seen by the agent is a `NamedTuple` containing the following:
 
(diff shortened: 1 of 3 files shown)


clement-bonnet:
Hi @raphaelavalos, good catch, thank you! Indeed, the grid is of type `int8`. I am happy with your fix. However, it would be helpful if we could fix the documentation at the same time.
- The `Cleaner` class docstring mentions `int32` instead of `int8` for the grid.
- The documentation readme at `docs/environments/cleaner.md` shows `int` instead of `int8`.
Would you please be able to bring these changes as well? 🙏

raphaelavalos:
Hi @clement-bonnet,
I included the docstring and docs modifications 😄

PR #220, opened 2024-01-24
feat(specs): make environment specs managed attributes

### LINKED ISSUE #98: feat(specs): make environment specs properties instead of methods
`env.observation_spec` is a method and creates a new spec each time it is being called. If we it a property, we can JIT a function that gets these specs.

```
def __init__(self):
    self.observation_spec = self._make_observation_spec()

def _make_observation_spec():
    return spec.Spec(...)
```

### LINKED ISSUE #221: feat(wrappers): Inherit type hints from environment
Currently, wrappers don't inherit type hinting from the environments they wrap. For example, consider the return types of the `reset` method. As you can see when using the environment directly, type hinting works as expected.
```
import jax

from jumanji.environments.logic.game_2048.env import Game2048, State
from jumanji.wrappers import Wrapper

key = jax.random.PRNGKey(0)

game_2048 = Game2048()
state, timestep = game_2048.reset(key)
reveal_type(state)  # jumanji.environments.logic.game_2048.types.State
reveal_type(
    timestep.observation
)  # Tuple[Any, Any, fallback=jumanji.environments.logic.game_2048.types.Observation]
```
However when using a wrapper around the environment, a couple problems show up. First, we are told that we need to annotate the wrapped environment. Second, type information about `state` and `timestep.observation` are lost.
```
game_2048 = Wrapper(Game2048())  # error: Need type annotation for "game_2048"
state, timestep = game_2048.reset(key)
reveal_type(state)  # Any
reveal_type(timestep.observation)  # Any
```
Explicitly annotating the environment fixes the error and restores the type hint on the `state` variable. However, there is still no type hint for `timestep.observation`.
```
game_2048: Wrapper[State] = Wrapper(Game2048())
state, timestep = game_2048.reset(key)
reveal_type(state)  # jumanji.environments.logic.game_2048.types.State
reveal_type(timestep.observation)  # Any
```
I believe a similar problem exists for most of the methods on the `Wrapper` class.
Ideally, the following code should just give proper type hints.
```
game_2048 = Wrapper(Game2048())
state, timestep = game_2048.reset(key)
reveal_type(state)  # jumanji.environments.logic.game_2048.types.State
reveal_type(
    timestep.observation
)  # Tuple[Any, Any, fallback=jumanji.environments.logic.game_2048.types.Observation]
```
The first issue of needing to explicitly annotate the wrapped environment can be solved by modifying the `__init__` method as follows
```
def __init__(self, env: Environment[State]):
    super().__init__()
    self._env = env
```
This will also provide type hints for the `state` variable. Providing type hints for `timestep.observation` will probably require adding a generic `Observation` typevar to the `Environment` class.

aar65537:
closes #98

- Implements private `_*_spec` properties on the Environment class. The properties are accessed through `*_spec`  managed attributes implemented with the `@property` decorator. Properties are initialized with `_make_*_spec()` methods during Environment initialization. 
- Implements generic `ActionSpec` type on the Environment class in order to statically type the `action_spec` attribute.
- Updates all environments to conform with changes to Environment class.
- Implements unit tests to ensure Environment specs can be accessed from a jitted function.
- Changes all references to `*_spec()` method to `*_spec` attribute.

Edit:
now also closes #221 

- Implements generic `Observation` type on `Environment`
- Implements generic types on Wrapper classes, so that Wrappers will inherit type hints from the Environments they wrap


jumanji/env.py
```diff
@@ -70,34 +78,68 @@ def step(self, state: State, action: chex.Array) -> Tuple[State, TimeStep]:
...
         """
+        return self._observation_spec
 
     @abc.abstractmethod
-    def action_spec(self) -> specs.Spec:
+    def _make_observation_spec(self) -> specs.Spec:
+        """Returns new observation spec.
+
+        Returns:
+            observation_spec: a NestedSpec tree of spec.
```
clement-bonnet, line 95: This is unrelated to this PR but all mentions of `NestedSpec` are wrong and come from legacy code. Could we possibly rename to:
```suggestion
            observation_spec: a potentially nested `Spec` structure representing the observation.
```

jumanji/env.py
```diff
@@ -33,9 +33,10 @@ class StateProtocol(Protocol):
 
 
 State = TypeVar("State", bound="StateProtocol")
+ActionSpec = TypeVar("ActionSpec", bound=specs.Array)
```
clement-bonnet, line 36: I think this adds unnecessary complexity to the code without serving a downstream purpose. Unless I'm mistaken, we don't seem to use `ActionSpec` for anything practical. Therefore, I would suggest removing the type variable and keeping the typing `spec.DiscreteArray` or whatever the environments have. What do you think?

clement-bonnet (changes requested):
Thank you @aar65537 for the contribution, it is a great improvement to the library! I just have one comment regarding the `ActionSpec` which I think we should not have.
Please let me know what you think!

jumanji/env.py
aar65537: Thank you. I was not aware of this. I will change the references to `NestedSpec`.

jumanji/env.py
aar65537: The purpose of `ActionSpec` is to type the `Environment.action_spec` method. Each environment class shouldn't need to override this method, so it needs to be typed correctly on the base `Environment` class. This requires a generic. The only other solution I can see would be to implement the following on each environment.
```
@property
def action_spec(self) -> specs.Whatever:
    return self._action_spec
```

clement-bonnet: Hmm I see. In this PR, it seems we are removing typing for the `observation_spec` method since we are moving it to the `Environment` base class. And we are doing similarly for the `action_spec`method but considering a `TypeVar` to keep some typing. Since the observation and action specs play similar roles as being environment-specific, I would argue for a similar implementation between the two.
- Either we define both `observation_spec` and `action_spec` in each environment class (allowing for environment-specific `spec.Spec[Observation]` and e.g. `spec.DiscreteArray` definitions).
- Or we just define the `make_observation_spec` and `make_action_spec` in each environment with a TypeVar-like typing.

I think solution 1 is clearer imo

aar65537: I agree that `observation_spec` and `action_spec` should probably have a similar implementation. However, I think solution 2 would be better. While it does lead to some unwieldy class definitions, it would also be a step towards addressing issue #221 with wrapper typing.

sash-a: I think I agree with @aar65537 on this. Personally prefer the TypeVar approach, as it allows us to not redefine `observation_spec` and `action_spec` in each environment. But we should definitely add a type for for the obs spec.

I have a couple concerns though. We may start having too many type vars in environment definitions e.g

```class Game2048(Environment[State, specs.DiscreteArray, specs.BoundedArray]):```

Which can be confusing to environment authors as to which generic refers to which spec, especially if we find the need to add more generics in the future.

Also are nested generics possible, could we ever do:

```class Game2048(Environment[State, specs.DiscreteArray, specs.Spec[Observation]]):```

I don't think they are, especially not in python 3.9/3.10. If this isn't possible I think we lose a lot more typing, so then I'd prefer the bloat of having both `observation_spec` and `make_observation_spec` methods in all classes for proper type definitions. 

A possible way around this is to have an `Observation` typevar and then type `observation_spec` in the base class as follows:

```
@property
def observation_spec() -> Specs.Spec[Observation]
```

Can we assume that this is always the type of observation spec?

(Thanks for this great contribution @aar65537! :fire:)

aar65537: I just updated the branch with the typevar approach and also fixed the issue with the wrappers. I share your concern with the number of typevars in the environment definitions, but I really don't see a better way.

Nested generics are valid, but they don't provide meaningful nesting. You can't access the inner type (`Observation` in this case) from the class. For this reason, I chose to use an `Observation` typevar and just assume `observation_spec` returns a `specs.Spec[Observation]`. This is true for all current environments, and in my opinion is a reasonable assumption. 

The only issue is some type information can be lost if your observation spec is a more specific type. For example, `FakeEnvironment` returns a `specs.Array` as its observation spec, but it is typed as `specs.Spec[chex.Array]`. `specs.Array` is a subtype of `specs.Spec[chex.Array]`, so there is no error, but the typing is not as precise as it could be.

sash-a: I think these assumptions are fine, I'm happy to merge this as is, @clement-bonnet what are your thoughts?

sash-a:
Sorry for the radio silence, I still want to get this in! What do you think about using cached_properties instead of properties as the spec should never change?

aar65537:
I think `cached_properties` should work. The only issue I foresee is you might have to access the properties in the `__init__` method to ensure the arrays are initialized outside of a jitted function.

sash-a:
> I think `cached_properties` should work. The only issue I foresee is you might have to access the properties in the `__init__` method to ensure the arrays are initialized outside of a jitted function.

True, but we know what the cached properties are, so we could call them in the base env classes `init`?

sash-a:
Also just a heads up, we're making a push to get sliding tile puzzle and flatpack envs merged before the ICLR camera ready deadline (about a week). Once this is done we'll get to this PR :smile:

aar65537:
Thanks for the heads up. I changed all of the environments to use `cached_property`. I think it is more succinct than my original implementation. Thank you for the suggestion.

sash-a (approved):
Thanks for this @aar65537, really nice change :smile:

aar65537:
I've implemented the changes for the SlidingTilePuzzle and cleaned up the last references to `action_spec()` in the docs. That should be all the environments.

PR #223, opened 2024-02-12
fix: autoreset wrappers

sash-a:
There was an issue with the autoreset wrappers: they never showed you the final timestep. This is an issue if the final timestep is a truncation (discount = 1, timestep.last() = true), then we'd want this observation in order to get the next value for our value target, but currently this would not be possible.

Gym fixes does it in the same way I am proposing as can be seen [here](https://github.com/vwxyzjn/cleanrl/blob/master/cleanrl/dqn.py#L195) they place the final observation in infos. What this PR does is always place either the current observation or the terminal observation in `timestep.extras["real_next_obs"]`

Here's how I'm currently using for SAC (and it's working well there):
```
    def step(
        action: Array, obs: Observation, env_state: State, buffer_state: BufferState
    ) -> Tuple[Array, State, BufferState, Dict]:
        """Given an action, step the environment and add to the buffer."""
        env_state, timestep = jax.vmap(env.step)(env_state, action)
        next_obs = timestep.observation
        rewards = timestep.reward
        terms = ~(timestep.discount).astype(bool)
        infos = timestep.extras

        real_next_obs = infos["real_next_obs"]

        transition = Transition(obs, action, rewards, terms, real_next_obs)
        buffer_state = rb.add(buffer_state, transition)

        return next_obs, env_state, buffer_state, infos["episode_metrics"]
```

### DIFF
--- a/jumanji/wrappers.py
+++ b/jumanji/wrappers.py
@@ -360,18 +360,40 @@ class VmapWrapper(Wrapper):
         return super().render(state_0)
 
 
+OBS_IN_EXTRAS_KEY = "real_next_obs"
+
+
+def _obs_in_extras(
+    state: State, timestep: TimeStep[Observation]
+) -> Tuple[State, TimeStep[Observation]]:
+    """Place the observation in timestep.extras[OBS_IN_EXTRAS_KEY].
+    Used when auto-resetting to store the observation from the terminal TimeStep.
+
+    Args:
+        state: State object containing the dynamics of the environment.
+        timestep: TimeStep object containing the timestep returned by the environment.
+
+    Returns:
+        (state, timestep): where the observation is placed in timestep.extras[OBS_IN_EXTRAS_KEY].
+    """
+    extras = timestep.extras
+    extras[OBS_IN_EXTRAS_KEY] = timestep.observation
+    return state, timestep.replace(extras=extras)  # type: ignore
+
+
 class AutoResetWrapper(Wrapper):
     """Automatically resets environments that are done. Once the terminal state is reached,
     the state, observation, and step_type are reset. The observation and step_type of the
     terminal TimeStep is reset to the reset observation and StepType.LAST, respectively.
     The reward, discount, and extras retrieved from the transition to the terminal state.
+    NOTE: The observation from the terminal TimeStep is stored in timestep.extras["real_next_obs"].
     WARNING: do not `jax.vmap` the wrapped environment (e.g. do not use with the `VmapWrapper`),
     which would lead to inefficient computation due to both the `step` and `reset` functions
     being processed each time `step` is called. Please use the `VmapAutoResetWrapper` instead.
     """
 
     def _auto_reset(
-        self, state: State, timestep: TimeStep
+        self, state: State, timestep: TimeStep[Observation]
     ) -> Tuple[State, TimeStep[Observation]]:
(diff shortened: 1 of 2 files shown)


jumanji/wrappers.py
```diff
@@ -360,18 +360,40 @@ def render(self, state: State) -> Any:
         return super().render(state_0)
 
 
+OBS_IN_EXTRAS_KEY = "real_next_obs"
+
+
+def _obs_in_extras(
```
sash-a, line 366: Not sure if this is the best place for the constant/method, but it's used only by the auto reset wrappers, so thought it made sense to put it above them :smile:

clement-bonnet:
Hi Sasha,
This issue seems a bit related to #106.
Returning the reset state/observation instead of the terminal state/observation when auto-resetting has always been the desired feature. This is because none of the Jumanji environments uses truncation, so one does not need the terminal state to train an actor-critic agent.
Now, if a user implements a new jumanji environments using the `Environment` abstraction and other tools from Jumanji, including truncation, one may want to use the truncated state/observation in their own training loop, which seems to be your use case, right? Passing it to the extras seems legit to me. 🙌

sash-a:
> Hi Sasha, This issue seems a bit related to #106. Returning the reset state/observation instead of the terminal state/observation when auto-resetting has always been the desired feature. This is because none of the Jumanji environments uses truncation, so one does not need the terminal state to train an actor-critic agent. Now, if a user implements a new jumanji environments using the `Environment` abstraction and other tools from Jumanji, including truncation, one may want to use the truncated state/observation in their own training loop, which seems to be your use case, right? Passing it to the extras seems legit to me. 🙌

Yup this is exactly the use case!

jumanji/wrappers.py
```diff
@@ -360,18 +360,40 @@ def render(self, state: State) -> Any:
         return super().render(state_0)
 
 
+OBS_IN_EXTRAS_KEY = "real_next_obs"
```
clement-bonnet, line 363: I think `NEXT_OBS_KEY_IN_EXTRAS` is probably preferred, and `"real_next_obs"` is ambiguous (because of "real"). What do you think of the following?
```suggestion
NEXT_OBS_KEY_IN_EXTRAS = "next_obs"
```

jumanji/wrappers.py
clement-bonnet: Having it here seems fine to me.

clement-bonnet (changes requested):

jumanji/wrappers.py
```diff
@@ -360,18 +360,40 @@ def render(self, state: State) -> Any:
...
+
+    Returns:
+        (state, timestep): where the observation is placed in timestep.extras[OBS_IN_EXTRAS_KEY].
+    """
+    extras = timestep.extras
+    extras[OBS_IN_EXTRAS_KEY] = timestep.observation
+    return state, timestep.replace(extras=extras)  # type: ignore
+
+
 class AutoResetWrapper(Wrapper):
```
clement-bonnet, line 384: We also need to apply the extras next obs to the `VmapAutoResetWrapper`

sash-a: This is already done?

clement-bonnet: Sorry, I didn't catch you applied it to both wrappers.

clement-bonnet (approved):

PR #224, opened 2024-02-14
fix: requirements

sash-a:
Sokoban introduced huggingface-hub as a requirement. It's only in requirements-train, but it's need to run jumanji, so moving it to the main requirements


clement-bonnet:
Good spot! LGTM, will look into why tests are failing.

clement-bonnet:
CI should be fixed after merging #225

clement-bonnet (approved):
LGTM

PR #225, opened 2024-03-04
ci: fix tests in pipeline

clement-bonnet:
- `jax.random.KeyArray` -> `jax.random.PRNGKey`
- use `jnp.isclose` instead of `==` for assertions with `float32`
- convert MMST's `frac` variable to `int32` because numpy uses `float64` by default and Jax does not.


sash-a (approved):
Thanks Clem

PR #246, opened 2024-09-23
Refactor: Minor code style improvements

helpingstar:
I have revised the variable assignment style and removed redundant code.

> As every PR requires CI testing, we discourage submitting PRs to fix one typo, one warning, etc. We recommend fixing the same issue at the file level at least (e.g., fix all typos in a file, fix all compiler warnings in a file, etc.)

I have read the above, but I wasn't sure if it applied to my case, so I submitted the PR for now. If this PR falls under the kind that should be avoided, I will close it.


sash-a (approved):
Thanks! :smile:

Issue #240, opened 2024-04-29
'docs/img/city_map.jpeg' not found during rendering


vmoens:
Calling render on "CVRP-v1" doesn't work because the docs/img/city_map.jpeg file cannot be opened:

https://github.com/instadeepai/jumanji/blob/823830e411b6e54da84985018555637d37c6bd8d/jumanji/environments/routing/tsp/viewer.py#L137

Perhaps a way of downloading that file in a cache could solve the issue?

clement-bonnet:
Hi, thank you for raising this issue! Will look into this shortly.

vmoens:
Cool!
FYI it's part of this PR
https://github.com/pytorch/rl/pull/2129
which enables rendering of envs using `from_pixels=True` in torchrl

Sort of personal request: It'd be nice to have a way to ask for headless rendering / rgb_arrays in a consistent way (from my 10K feet high view it seems envs have a different way of dealing with rendering). Something like `make(envname, rgb_array=True)` or `render(rgb_array=True)`.
For torchrl, we prefer always getting arrays of uint8 and deliver them to the users as part of their env output and leave them deal with them as they want (if the user wants to render smth they can just call `plt.imshow(data["pixels"])`)
Having the human rendering by default usually introduces some consequent overhead.

The way the above PR deals with this is to render on screen, record the content of the plt image in a buffer, convert that buffer to a tensor (since there is no generic way of getting a rgb).

sash-a:
Sorry it took so long @vmoens but this should be fixed now :smile:

PR #251, opened 2024-10-25
fix: rendering in cvrp and tsp

sash-a:
Adds the `city_map.jpeg` to all environments that use it, following the same style as sokoban.

Closes: #240


WiemKhlifi (approved):
Thanks @sash-a 🛥️

Issue #248, opened 2024-10-22
Create a naming convention for Jumanji envs


WiemKhlifi:
### Is your feature request related to a problem? Please describe
Configuring environments in **Jumanji** involves manually setting different parameters such as grid size, number of agents etc. This manual setup is not only time consuming but complicates the integration of Jumanji envs with RL frameworks like **MAVA**. Each new scenario addition currently requires a distinct YAML configuration, as seen [here in the MAVA repo](https://github.com/instadeepai/Mava/tree/develop/mava/configs/env/scenario).

### Describe the solution you'd like
I propose implementing a standardised naming convention for Jumanji envs that succinctly encodes all necessary parameters into the environment's name. This approach would mimic the simplicity of classical Gym envs. For example:
```python
# Initialize the Level-Based Foraging environment with concise identifiers.
env = gym.make("Foraging-2s-8x8-2p-2f-coop-v3")
```

A function to extract and apply this naming convention could look like this:
```python
def get_lbf_config(scenario):
    # Example format: "2s-10x10-3p-3f-coop"
    parts = scenario.split('-')
    grid_size = int(parts[1].split('x')[0])
    fov = int(parts[0].rstrip('s')) if 's' in parts[0] else grid_size
    num_agents = int(parts[2].rstrip('p'))
    num_food = int(parts[3].rstrip('f'))
    force_coop = 'coop' in parts

    return {
        "grid_size": grid_size,
        "fov": fov,
        "num_agents": num_agents,
        "num_food": num_food,
        "force_coop": force_coop,
        "max_agent_level": 2,  # Set as default
    }
```

### Describe alternatives you've considered
An alternative could involve establishing a registration system akin to Gym's, where we can register environments with predefined attributes:

```python
from gymnasium import register

# Example permutations for environment registration
for s, p, f, coop, po in itertools.product(
    range(5, 20), range(2, 10), range(1, 10), [True, False], [True, False]
):
    register(
        id=f"Foraging-{'2s' if po else ''}{s}x{s}-{p}p-{f}f{'-coop' if coop else ''}-v3",
        entry_point="lbforaging.foraging:ForagingEnv",
        kwargs={
            "field_size": (s, s),
            "num_agents": p,
            "num_food": f,
            "sight": 2 if po else s,
            "force_coop": coop,
            "max_episode_steps": 50,
            "grid_observation": not po,
        },
    )
```

---
### Misc
- [x] Check for duplicate requests.

sash-a:
[...]

PR #218, opened 2024-01-18
Feat: Full Level-Based Foraging(LBF) environment

WiemKhlifi:
In this PR, we have introduced an implementation of the [Level-Based Foraging (LBF)](https://github.com/semitable/lb-foraging) environment, incorporating actor-critic and random networks along with documentation.

Note: The gym version of the original implementation of [LBF](https://github.com/semitable/lb-foraging).

PR #259, opened 2024-11-04
Implement Search & Rescue Multi-Agent Environment

zombie-einstein:
Add a multi-agent search and rescue environment where a set of agents has to locate moving targets on a 2d space.

## Changes

- Adds the [Esquilax library](https://github.com/zombie-einstein/esquilax) as a dependency
- Adds the `swarm` environment group/type (was not sure the new environment fit into an existing group, but happy to move if you think it would better fit somewhere else)
- Implement some common swarm/flock functionality (can be used if more environments of this type are added)
- Implement the search and rescue environment and docs

## Todo

- Need to add images and animations, was waiting to finalise code before adding.


zombie-einstein:
Here you go @sash-a this is correct now. Will grab a look at  the contributor license and Ci failure now.

zombie-einstein:
I think CI issue is I've Esquilax set to Python `>=3.10`, seems you've a PR open to upgrade Python version, is it worth holding on for that?

sash-a:
Python version PR is merged now so hopefully it will pass :smile: 

Should have time during the week to review this, really appreciate the contribution!

sash-a (changes requested):
An initial review with some high level comments about jumanji conventions. Will go through it more in depth once these are addressed. In general it's looking really nice and well documented! 

Not quite sure on the new swarms package, but also not sure where else we would put it. Not sure on it especially if we only have 1 env and no news ones planned. 

One thing I don't quite understand is the benefit of `amap` over `vmap` specifically in the case of this env?

Please @ me when it's ready for another review or if you have any questions.

sash-a:
As for your questions in the description:
> I only forwarded the Environment import to jumanji.environments do types also need forwarding somewhere?

Nope just the environment is fine

> I didn't add an animate method to the environment, but saw that some other do? Easy enough to add.

Please do add animation it's a great help.

> Do you want defaults for all the environment parameters? Not sure there are really "natural" choices, but could add sensible defaults to avoid some typing.

We do want defaults, I think we can discuss what makes sense.

> Are the API docs auto-generated somehow, or do I need to add a link manually?

It's generated with mkdocs, we need an entry in `docs/api/environments` and `mkdocs.yml`, see [this recently closed PR](https://github.com/instadeepai/jumanji/pull/218/files) for an example of which files we change

One big thing I've realized that this is missing after my review is training code. We like to validate that the env works. I'm not 100% sure if this is possible because the env has two teams, so which reward do you optimize, maybe training with simple heuristic, eg you are the predator and the prey moves randomly? For examples see the `training` folder, you should only need to create a network. An example of this should also be in the above PR.

jumanji/environments/swarms/common/types.py
```diff
@@ -0,0 +1,53 @@
...
+
+if TYPE_CHECKING:
+    from dataclasses import dataclass
+else:
+    from chex import dataclass
+
+import chex
+
+
+@dataclass
```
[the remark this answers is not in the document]
zombie-einstein: Correct yeah these remain fixed after creation, will add.

jumanji/environments/swarms/common/updates.py
```diff
@@ -0,0 +1,183 @@
...
+    new_speeds = jnp.clip(
+        boid.speed + acceleration,
+        min=params.min_speed,
+        max=params.max_speed,
+    )
+
+    return new_heading, new_speeds
+
+
+@esquilax.transforms.amap
```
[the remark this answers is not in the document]
zombie-einstein: Yes in this case this is overkill, I'll use vmap inside the update function.

jumanji/environments/swarms/common/updates.py
```diff
@@ -0,0 +1,183 @@
...
+
+from typing import Tuple
+
+import chex
+import esquilax
+import jax
+import jax.numpy as jnp
+
+from . import types
+from .types import AgentParams
```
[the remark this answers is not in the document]
zombie-einstein: I'll switch out relative imports (there's a few around). In this case also have `types.Agentstate` so will import the types module.

jumanji/environments/swarms/common/updates.py
```diff
@@ -0,0 +1,183 @@
...
+
+    return types.AgentState(
+        pos=positions,
+        speed=speeds,
+        heading=headings,
+    )
+
+
+def view(
+    _: chex.PRNGKey,
```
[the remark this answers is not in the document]
zombie-einstein: Yeah Esquilax expects mapped functions to have a certain signature, though for cases like this it would be nice to be able to omit. I've not really settled on a clean way of having a function that I can optionally pass in arguments, but it is on y radar.
I'll add something in the docstring to explain.

jumanji/environments/swarms/predator_prey/types.py
```diff
@@ -0,0 +1,71 @@
...
+
+@dataclass
+class Rewards:
+    """
+    predators: Array of individual rewards for predator agents.
+    prey: Array of individual rewards for prey agents.
+    """
+
+    predators: chex.Array
+    prey: chex.Array
```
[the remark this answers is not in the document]
zombie-einstein: Yeah I did this in the prototype to indicate something that just had the two fields. I guess you could say for readability and in the strict typing sense these should be different things (was my thinking here)? But also appreciate the repetition is a bit ugly.

jumanji/environments/swarms/common/types.py
```diff
@@ -0,0 +1,53 @@
...
+    State of multiple agents of a single type
+
+    pos: 2d position of the (centre of the) agents
+    heading: Heading of the agents (in radians)
+    speed: Speed of the agents
+    """
+
+    pos: chex.Array
+    heading: chex.Array
+    speed: chex.Array
```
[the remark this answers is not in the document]
zombie-einstein: Oh yes I'll add in

jumanji/environments/swarms/predator_prey/updates.py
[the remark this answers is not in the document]
zombie-einstein: Oh nice yeah, I meant to ask about generic reward functions, reward tuning can be a large part of these multi-agent environments

jumanji/environments/swarms/predator_prey/types.py
```diff
@@ -0,0 +1,71 @@
...
+
+@dataclass
+class Observation:
+    """
+    predators: Local view of predator agents.
+    prey: Local view of prey agents.
+    """
+
+    predators: chex.Array
+    prey: chex.Array
```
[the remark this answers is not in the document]
zombie-einstein: What's the reason for this (for my own knowledge)?

zombie-einstein:
Hi @sash-a, just merged changes that I think  address all the comments, and the `animate` method, and API docs link.

> Not quite sure on the new swarms package, but also not sure where else we would put it. Not sure on it especially if we only have 1 env and no news ones planned.

Could you have something like a `multi-agent` package? Don't think you have similar at the moment? FYI was intending to add a couple more swarm/flock type envs if this one went ok.

> One thing I don't quite understand is the benefit of amap over vmap specifically in the case of this env? 

Yeah in a couple cases using it is overkill, hang-over from when I was writing this example with esquilax demo in mind! Makes sense to use `vmap` instead if the other arguments are not being used.

zombie-einstein:
I'll look at adding something to training next. I think random prey with trained predators makes sense, will look to implement.

jumanji/environments/swarms/predator_prey/types.py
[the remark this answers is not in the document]
sash-a: Just a legacy thing once it's jitted I think there's very little difference between named tuple and dataclass

jumanji/environments/swarms/predator_prey/types.py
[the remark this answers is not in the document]
sash-a: It's not a big deal, but I would just make a `Pair` or `PredatorPrey` class for this

sash-a:
One other thing, the only reason I've been hesitant to add this to Jumanji is because it's not that related to industry problems which is a common focus between all the envs. I was thinking maybe we could re-frame the env from predator-prey to something else (without changing any code, just changing the idea). I was thinking maybe a continuous cleaner where your target position is changing or something to do with drones, do you have any other ideas and would you be happy with this?

zombie-einstein:
> > Could you have something like a multi-agent package? Don't think you have similar at the moment? FYI was intending to add a couple more swarm/flock type envs if this one went ok.
> 
> If you can add more that would be great! Then I'm happy to keep the swarm package as is. What we'd be most interested in is some kind of env with only 1 team and strictly co-operative like predators vs heuristic prey or visa versa, not sure if you planned to make any envs like this?

Yeah I was very interested in developing envs for co-operative multi-agent RL so was keen to design or implement more environments along theses lines. There's a simpler version of this environment which is just the flock, i.e. where the agents move in a co-ordinated way with out colliding. Also seen an environment where the agents have to effectively cover an an area that I was going to look at.
 
> Also I updated the CI yesterday, we're now using ruff, so you will need to update your pre-commit

How do I do this? I did try reinstalling pre-commit, but it raised an error that the config was invalid?

zombie-einstein:
> One other thing, the only reason I've been hesitant to add this to Jumanji is because it's not that related to industry problems which is a common focus between all the envs. I was thinking maybe we could re-frame the env from predator-prey to something else (without changing any code, just changing the idea). I was thinking maybe a continuous cleaner where your target position is changing or something to do with drones (maybe delivery), do you have any other ideas and would you be happy with this?

Yeah definitely open to suggestions. I was thinking more in the abstract for this (will the agents develop some collective behaviour to avoid predators) but happy to modify towards something more concrete.

sash-a:
Great to hear on the co-operative marl front those both sound like nice envs to have

> How do I do this? I did try reinstalling pre-commit, but it raised an error that the config was invalid?

Couple things to try:
```
pip install -U pre-commit
pre-commit uninstall
pre-commit install
```

If this doesn't work check `which pre-commit` it should point to your virtual environment if it's pointing to your system python or some other system folder just uninstall that version and rerun the above.

> Yeah definitely open to suggestions. I was thinking more in the abstract for this (will the agents develop some collective behaviour to avoid predators) but happy to modify towards something more concrete.

Agreed it would be nice to keep it abstract for the sake of research, but I think it's nice that this env suite is all industry focused. I quite like something to do with drones - seems quite industry focused although we must definitely avoid anything to do with war. I'll give it a think

zombie-einstein:
Hi @sash-a fixed the formatting and consolidated the predator-prey type.

sash-a:
Thanks I'll try have a look tomorrow, sorry previous 2 days were a bit more busy than expected. 

For the theme I'm think maratime search and rescue works well. It's relatively real world and fits the current dynamics

zombie-einstein:
> Thanks I'll try have a look tomorrow, sorry previous 2 days were a bit more busy than expected.
> 
> For the theme I'm think maratime search and rescue works well. It's relatively real world and fits the current dynamic

Thanks, no worries. Actually yeah funnily enough a co-ordinated search was something I'd been looking into. Yeah could have one set of agent have some drift w random movements that need to be found inside the simulated region.

sash-a:
Sorry still didn't have time to review today and Mondays are usually super busy for me, but I'll get to this next week!

As for the theme do you think we should then change the dynamics a bit to make prey heuristically controlled to move sort of randomly?

zombie-einstein:
> Sorry still didn't have time to review today and Mondays are usually super busy for me, but I'll get to this next week!
> 
> As for the theme do you think we should then change the dynamics a bit to make prey heuristically controlled to move sort of randomly?

No worries, sure I'll do a revision this weekend!

zombie-einstein:
Hi @sash-a, this turned into a larger rewrite (sorry for the extra review work, let me know if you want me to close this PR and just start with a fresh one) but think it's a more realistic scenario

- A team of agents is searching for targets in the environment region
- Targets are controlled by an fixed update algorithm (that has an interface to allow other behaviours)
- Agents are only rewarded the first time a target is located
- To detect the target they must come within a fixed range of them.
- Agents visualise the local environment, i.e. the location of of other agents in their vicinity.

A couple choices we may want to consider:

- Agents are individually rewarded, we could have some interface for reward shaping (to promote co-operation), but could also leave this external to the environment for the user to implement?
- At the moment agents only visualise other neighbours. A twist on this I considered was once targets are revealed they are then visualised (i.e. can be seen) by each agent as part of their local view.
- Do we want to scale rewards with how quickly targets are found, feels like it would make sense?
- I've assigned a fixed number of steps to locate the targets, but also seems it would makes sense to terminate the episode when all located?
- As part of the observation I've included the remaining steps and targets as normalised floats, but not sure if you have some convention for values like this (i.e. just use integer values and let use rescale them)

sash-a:
Thanks for this @zombie-einstein I'll start having a look now :smile:  
I think leave the PR as is, no need to create a new one.

> that has an interface to allow other behaviors

awesome!


> Agents are only rewarded the first time a target is located

Agreed I think we should actually hide targets once they are located so as to not confuse other agents. 


> Agents are individually rewarded

I think individual is fine and externally users can sum it outside if they want. e.g we do this in [mava for connector](https://github.com/instadeepai/Mava/blob/develop/mava/wrappers/jumanji.py#L254)
 
> At the moment agents only visualise other neighbours. A twist on this I considered was once targets are revealed they are then visualised (i.e. can be seen) by each agent as part of their local view.

Not quite following what you mean here. I would say an agent should observe all agents and targets (that have not yet been rescued) within their local view.

> Do we want to scale rewards with how quickly targets are found, feels like it would make sense?

Maybe add this as an optional reward type, I think I prefer 1 if target is saved and 0 otherwise - makes the env quite hard, but we should test what works best.

> I've assigned a fixed number of steps to locate the targets, but also seems it would makes sense to terminate the episode when all located?

Definitely!
    
> As part of the observation I've included the remaining steps and targets as normalised floats, but not sure if you have some convention for values like this (i.e. just use integer values and let use rescale them)

We don't have a convention for this. I wouldn't add remaining steps to the obs directly I don't see why the algorithm would need that, although again needs to be tested. Agreed with remaining targets, makes sense to observe that. I think normalised floats makes sense.

docs/environments/search_and_rescue.md
```diff
@@ -0,0 +1,54 @@
...
+  For example if an agent sees another agent straight ahead and `num_vision = 5` then
+  the observation array could be
+
+  ```
+  [1.0, 1.0, 0.5, 1.0, 1.0]
+  ```
+
+  where `1.0` indicates there is no agents along that ray, and `0.5` is the normalised
+  distance to the other agent.
+- `target_remaining`: float in the range [0, 1]. The normalised number of targets
```
sash-a, line 38: minor grammar nitpick can we call this target**s**_remaining

docs/environments/search_and_rescue.md
```diff
@@ -0,0 +1,54 @@
...
+  along a ray from the agent to the nearest neighbour, with  each cell representing a
+  ray angle (with `num_vision` rays evenly distributed over the agents field of vision).
+  For example if an agent sees another agent straight ahead and `num_vision = 5` then
+  the observation array could be
+
+  ```
+  [1.0, 1.0, 0.5, 1.0, 1.0]
+  ```
+
+  where `1.0` indicates there is no agents along that ray, and `0.5` is the normalised
```
sash-a, line 36: I feel either 0 or -1 should indicate there is no agent along the ray, what do you think? Feels more natural to me.

Although I guess 1 indicates out of range, so 0 would be right in front. So maybe -1 is best...not sure what you think?

docs/environments/search_and_rescue.md
```diff
@@ -0,0 +1,54 @@
...
+  distance to the other agent.
+- `target_remaining`: float in the range [0, 1]. The normalised number of targets
+  remaining to be detected (i.e. 1.0 when no targets have been found).
+- `time_remaining`: float in the range [0, 1]. The normalised number of steps remaining
```
sash-a, line 40: I would make this step count instead similar to other envs

docs/environments/search_and_rescue.md
```diff
@@ -0,0 +1,54 @@
...
+  to locate the targets (i.e. 0.0 at the end of the episode).
+
+## Actions
+
+Jax array (float) of `(num_searchers, 2)` in the range [-1, 1]. Each entry in the
```
sash-a, line 45:
```suggestion
Jax array (float) of `(num_searchers, 2)` in the range `[-1, 1]`. Each entry in the
```

docs/environments/search_and_rescue.md
sash-a, line 40:
```suggestion
- `time_remaining`: float in the range `[0, 1]`. The normalised number of steps remaining
```

docs/environments/search_and_rescue.md
sash-a, line 38:
```suggestion
- `target_remaining`: float in the range `[0, 1]`. The normalised number of targets
```

docs/environments/search_and_rescue.md
```diff
@@ -0,0 +1,54 @@
...
+array represents an update of each agents velocity in the next step. Searching agents
+update their velocity each step by  rotating and accelerating/decelerating. Values
+are clipped to the range `[-1, 1]` and then scaled by max rotation and acceleration
+parameters. Agents are restricted to velocities within a fixed range of speeds.
```
sash-a, line 49: Only 1 thing missing here can you explain which of the two values is acceleration and which is deceleration. Also could you explain what 1 and -1 mean for rotation, is it angular speed or an immediate rotation?

docs/environments/search_and_rescue.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Rewards
+
+Jax array (float) of `(num_searchers, 2)`. Rewards are generated for each agent individually.
```
sash-a, line 53: Just a single reward per agent, not 2 right?
```suggestion
Jax array (float) of `(num_searchers,)`. Rewards are generated for each agent individually.
```

jumanji/environments/swarms/search_and_rescue/dynamics.py
```diff
@@ -0,0 +1,61 @@
...
+        Args:
+            key: random key.
+            target_pos: Current target positions.
+
+        Returns:
+            Updated target positions.
+        """
+        d_pos = jax.random.uniform(key, target_pos.shape)
+        d_pos = self.step_size * 2.0 * (d_pos - 0.5)
+        return target_pos + d_pos
```
sash-a, line 61: Worried this might make their movement quite "jerky" have you tried visualizing it? I would say updating their velocity/angular velocity would make them move a bit smoother?

sash-a (changes requested):
Amazing job with this rewrite, haven't had time to fully look at everything but it does look great so far!

Some high level things:
* Please add a generator, dynamics and viewer test (see examples of the viewer test for other envs)
* Can you also add tests for the common/updates
* Can you start looking into the networks and testing for jumanji

Sorry a bit tedious tasks, but I really like the env we've landed on :smile:

docs/environments/search_and_rescue.md
zombie-einstein: Yeah my thinking was that 1.0 means objects/agents move away they do so continuosly until they are out of range at 0, felt the jump to -1 would maybe introduce slightly unatural discontinuity?

docs/environments/search_and_rescue.md
zombie-einstein: Added a longer explanation here

jumanji/environments/swarms/common/updates.py
```diff
@@ -0,0 +1,178 @@
...
+
+    positions = jax.random.uniform(k1, (n, 2))
+    speeds = jax.random.uniform(k2, (n,), minval=params.min_speed, maxval=params.max_speed)
+    headings = jax.random.uniform(k3, (n,), minval=0.0, maxval=2.0 * jnp.pi)
+
+    return types.AgentState(
+        pos=positions,
+        speed=speeds,
+        heading=headings,
+    )
```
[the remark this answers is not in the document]
zombie-einstein: As well as this I've the agent related types in this `common` module in anticipation of usage in other environments. Want to consoildate the whole thing?

jumanji/environments/swarms/search_and_rescue/env.py
```diff
@@ -0,0 +1,429 @@
...
+) -> chex.Array:
+    """
+    Return +1.0 reward if the target is within the searcher view angle.
+    """
+    dx = shortest_vector(searcher.pos, target.pos)
+    phi = jnp.arctan2(dx[1], dx[0]) % (2 * jnp.pi)
+    dh = shortest_vector(phi, searcher.heading, 2 * jnp.pi)
+    can_see = (dh >= -searcher_view_angle) & (dh <= searcher_view_angle)
+    return jax.lax.cond(
+        # (not target.found) & can_see,
```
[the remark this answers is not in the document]
zombie-einstein: Thanks, yeah this is a mistake. Commented as I was having some type issues, but yeah think `not` was the issue.

jumanji/environments/swarms/search_and_rescue/env.py
```diff
@@ -0,0 +1,429 @@
...
+            max_accelerate=searcher_max_accelerate,
+            min_speed=searcher_min_speed,
+            max_speed=searcher_max_speed,
+            view_angle=searcher_view_angle,
+        )
+        self.max_steps = max_steps
+        super().__init__()
+        self._viewer = viewer or SearchAndRescueViewer()
+        self._target_dynamics = target_dynamics or RandomWalk(0.01)
+        self._generator = generator or RandomGenerator(num_searchers, num_targets)
```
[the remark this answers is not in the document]
zombie-einstein: Oh yeah good point!

zombie-einstein:
Thanks @sash-a, just a couple follow ups to your questions:

> > At the moment agents only visualise other neighbours. A twist on this I considered was once targets are revealed they are then visualised (i.e. can be seen) by each agent as part of their local view.
> 
> Not quite following what you mean here. I would say an agent should observe all agents and targets (that have not yet been rescued) within their local view.
> 

So I was picturing (and as currently implemented) a situation where the searchers have to come quite close the targets to "find" them (as if they are obscured/hard to find), but the agents have a larger vision range to visualise the location of other searchers agents (to allow them to improve search patterns for example). 

My feeling was that this created more of a search task, where if the targets are part of their larger vision range it feels like it could be more of a routing type task. 

I then thought it may be good to include found targets in the vision to allow agents to visualise density of located targets.

> > As part of the observation I've included the remaining steps and targets as normalised floats, but not sure if you have some convention for values like this (i.e. just use integer values and let use rescale them)
> 
> We don't have a convention for this. I wouldn't add remaining steps to the obs directly I don't see why the algorithm would need that, although again needs to be tested. Agreed with remaining targets, makes sense to observe that. I think normalised floats makes sense.

I thought if treating it as a time-sensitive task some indication of the remaining time to find targets could be a useful feature of the observation.

>  Please add a generator, dynamics and viewer test (see examples of the viewer test for other envs)
    Can you also add tests for the common/updates
    Can you start looking into the networks and testing for jumanji

Yup will do!

jumanji/environments/swarms/search_and_rescue/dynamics.py
zombie-einstein: Yeah fair point, was thinking about them being something like objects in the ocean with random motion (and maybe some overall drift) but as is this could also be quite restrictive given we only track the position of targets. 
So guess it makes sense to include velocity as part of the target state, and tweak the default behaviour to be random velocity changes?

sash-a:
Hi @zombie-einstein 

> So I was picturing (and as currently implemented) a situation where the searchers have to come quite close the targets to "find" them (as if they are obscured/hard to find), but the agents have a larger vision range to visualise the location of other searchers agents (to allow them to improve search patterns for example).

I think this is great!

> I then thought it may be good to include found targets in the vision to allow agents to visualise density of located targets.

I see the thinking here, but I'm not even sure it's that beneficial because targets can move right?

One thing I'm a bit concerned about, as you increase the number of agents will the problem not get easier? As I don't see an option to increase the world size, so as the number of agents increases the density of searchers increases making it easier to find targets. Is there a way we could increase the world size or another way to avoid this issue?

Sorry been quite busy this last week, but I should be a lot more time next week to dedicate more time to this review :smile:

zombie-einstein:
Hey @sash-a, no worries, still got stuff to get on  with.

> > I then thought it may be good to include found targets in the vision to allow agents to visualise density of located targets.
> 
> I see the thinking here, but I'm not even sure it's that beneficial because targets can move right?
>

Yeah this is correct. I guess it depends on the target dynamics. For something simple like noisey movement with some drift it could help with identifying drift and areas of low density that have not been searched yet? Kind of like the agents are communicating what they've found/have some memory. 
 
> One thing I'm a bit concerned about, as you increase the number of agents will the problem not get easier? As I don't see an option to increase the world size, so as the number of agents increases the density of searchers increases making it easier to find targets. Is there a way we could increase the world size or another way to avoid this issue?
> 

Yeah it would. The region is fixed in Esquilax to the unit square, mainly just reduce the number of parameters used in describing the interaction between agents (and make my life a bit easier 😂) but could be something to add into the library. The way to avoid this here would be scaling other parameters, e.g. scaling the vision range and speed range of agents, the only issue possibly being numerical accuracy. 

For the network, is there a built-in way to do multi-agent training? If not I guess the most straightforward way to get it working would be to just have a single agent, and just wrap some means of flattening the rewards?

sash-a:
> Yeah this is correct. I guess it depends on the target dynamics.

True, maybe we can have it be an option? Although might get messy to define observation shapes

> The region is fixed in Esquilax to the unit square

I see, so how many agents do you think it could scale to before it gets too crowded or too numerically unstable?

> For the network, is there a built-in way to do multi-agent training? If not I guess the most straightforward way to get it working would be to just have a single agent, and just wrap some means of flattening the rewards?

That's exactly what we do. We wrap things in a the [multi to single wrapper](https://github.com/instadeepai/jumanji/blob/main/jumanji/wrappers.py#L221) and then treat it as a single agent problem. See how we do the learning for [LBF](https://github.com/instadeepai/jumanji/blob/main/jumanji/training/networks/lbf/actor_critic.py) and [Connector](https://github.com/instadeepai/jumanji/blob/main/jumanji/training/networks/connector/actor_critic.py) and this [if statement](https://github.com/instadeepai/jumanji/blob/main/jumanji/training/setup_train.py#L95) in the trainer setup.

jumanji/environments/swarms/search_and_rescue/test_utils.py
```diff
@@ -0,0 +1,107 @@
...
+
+    state = generator(key, params)
+
+    assert isinstance(state, State)
+    assert state.searchers.pos.shape == (generator.num_searchers, 2)
+    assert jnp.all(0.0 <= state.searchers.pos) and jnp.all(state.searchers.pos <= 1.0)
+    assert state.targets.pos.shape == (generator.num_targets, 2)
+    assert jnp.all(0.0 <= state.targets.pos) and jnp.all(state.targets.pos <= 1.0)
+    assert not jnp.any(state.targets.found)
+    assert state.step == 0
```
sash-a, line 49: Can you put this in the `test_generator.py`

jumanji/environments/swarms/search_and_rescue/types.py
```diff
@@ -0,0 +1,71 @@
...
+else:
+    from chex import dataclass
+
+import chex
+
+from jumanji.environments.swarms.common.types import AgentState
+
+
+@dataclass
+class TargetState:
```
sash-a, line 27:
```suggestion
class TargetState:
"""The state for the rescue targets."""
```

jumanji/environments/swarms/search_and_rescue/utils.py
```diff
@@ -0,0 +1,88 @@
...
+            search status).
+
+    Returns:
+        reward: +1.0 reward if the agent detects a new target.
+    """
+    dx = shortest_vector(searcher.pos, target.pos)
+    phi = jnp.arctan2(dx[1], dx[0]) % (2 * jnp.pi)
+    dh = shortest_vector(phi, searcher.heading, 2 * jnp.pi)
+    searcher_view_angle = searcher_view_angle * jnp.pi
+    can_see = (dh >= -searcher_view_angle) & (dh <= searcher_view_angle)
```
sash-a, line 83: This seems like it's the same as the code above so you can probably just re-use that method?

jumanji/environments/swarms/search_and_rescue/utils.py
```diff
@@ -0,0 +1,88 @@
...
+import chex
+import jax
+import jax.numpy as jnp
+from esquilax.utils import shortest_vector
+
+from jumanji.environments.swarms.common.types import AgentState
+from jumanji.environments.swarms.search_and_rescue.types import TargetState
+
+
+def has_been_found(
```
sash-a, line 24: These method names are very similar and quite confusing.

I'd call one `target_in_range` and the other `reward_if_found_target`

jumanji/environments/swarms/search_and_rescue/utils.py
```diff
@@ -0,0 +1,88 @@
...
+    dx = shortest_vector(searcher.pos, target.pos)
+    phi = jnp.arctan2(dx[1], dx[0]) % (2 * jnp.pi)
+    dh = shortest_vector(phi, searcher.heading, 2 * jnp.pi)
+    searcher_view_angle = searcher_view_angle * jnp.pi
+    can_see = (dh >= -searcher_view_angle) & (dh <= searcher_view_angle)
+    return jax.lax.cond(
+        ~target.found & can_see,
+        lambda: 1.0,
+        lambda: 0.0,
+    )
```
sash-a, line 88: In general it's best to avoid `cond` if at all possible as it may call to the CPU for some conditions, here I think it's fine to just do the following:
```suggestion
    (~target.found & can_see).astype(float)
```

jumanji/environments/swarms/search_and_rescue/env.py
```diff
@@ -0,0 +1,390 @@
...
+        self,
+        searcher_vision_range: float,
+        target_contact_range: float,
+        num_vision: int,
+        agent_radius: float,
+        searcher_max_rotate: float,
+        searcher_max_accelerate: float,
+        searcher_min_speed: float,
+        searcher_max_speed: float,
+        searcher_view_angle: float,
```
sash-a, line 114: Need to decide on sensible defaults for all of these

jumanji/environments/swarms/search_and_rescue/env.py
```diff
@@ -0,0 +1,390 @@
...
+        Args:
+            state: Environment state.
+            actions: Arrays of searcher steering actions.
+
+        Returns:
+            state: Updated searcher and target positions and velocities.
+            timestep: Transition timestep with individual agent local observations.
+        """
+        searchers = update_state(state.key, self.searcher_params, state.searchers, actions)
+        key, target_key = jax.random.split(state.key, num=2)
```
sash-a, line 220: Looks like you're re-using `key` quite a few times, be sure to split up here for as many times as you need random calls. Also I'd recommend doing this at the top of the function and then pass one of the split keys into `update_state`

jumanji/environments/swarms/search_and_rescue/env.py
```diff
@@ -0,0 +1,390 @@
...
+            step=state.step + 1,
+        )
+        observation = self._state_to_observation(state)
+        timestep = jax.lax.cond(
+            state.step >= self.max_steps | jnp.all(targets_found),
+            termination,
+            transition,
+            rewards,
+            observation,
+        )
```
sash-a, line 270: Just wondering if you need to specify the `shape` parameter here as it might not return the correct shapes otherwise. Although this may only matter for the `discount`

jumanji/environments/swarms/search_and_rescue/env.py
```diff
@@ -0,0 +1,390 @@
...
+            searchers,
+            state.targets,
+            pos=searchers.pos,
+            pos_b=target_pos,
+        )
+        # Mark targets as found if with contact range and view angle of a searcher
+        # spatial maps the has_been_found function over all pair of targets and
+        # searchers within range of each other
+        targets_found = spatial(
+            utils.has_been_found,
```
sash-a, line 243: With these two function it seems like you're doing the same computation twice, I'd recommend checking which targets are found this step and then passing that to your reward function (which you could then move out of utils and put in here because it should be quite simple)

jumanji/environments/swarms/search_and_rescue/env.py
```diff
@@ -0,0 +1,390 @@
...
+        targets_found = jnp.logical_or(targets_found, state.targets.found)
+        state = State(
+            searchers=searchers,
+            targets=TargetState(pos=target_pos, found=targets_found),
+            key=key,
+            step=state.step + 1,
+        )
+        observation = self._state_to_observation(state)
+        timestep = jax.lax.cond(
+            state.step >= self.max_steps | jnp.all(targets_found),
```
sash-a, line 265: In general it's good to be quite overly careful with the brackets for jax conditions, I think this may have done an or with `max_steps` and `targets_found`
```suggestion
            (state.step >= self.max_steps) | jnp.all(targets_found),
```

sash-a (commented):
Finally had time to go through this, it's looking really great. Some optimization suggestions and minor nitpicks.

I want to give the documentation a once over, but once you've addressed all these and added networks so we can see learning curves I'm happy to merge this :smile:

jumanji/environments/swarms/common/updates.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def view(
+    _key: chex.PRNGKey,
+    params: Tuple[float, float],
```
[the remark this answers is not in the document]
zombie-einstein: This is a design choice in Esquilax, how broadcasting/mapping of arguments is decided is done positionally (so in this case the params are broadcast rather than mapped across the mapped interactions). 
It could be nice to have some alternate way to mark how arguments are to be used, but not settled on anything clean yet.

jumanji/environments/swarms/search_and_rescue/env_test.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+def test_env_step(env: SearchAndRescue) -> None:
+    """
+    Run several steps of the environment with random actions and
+    check states (i.e. positions, heading, speeds) all fall
+    inside expected ranges.
+    """
+    key = jax.random.PRNGKey(101)
+    n_steps = 22
```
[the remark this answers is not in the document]
zombie-einstein: Not a particular one, mainly to have a distinct number when shape checking, but also something largish that will mean the random actions cover the action/state space sufficiently (e.g catching thing like looping at the boundaries).

jumanji/environments/swarms/search_and_rescue/utils.py
zombie-einstein: Yeah I can pull out the common code, but then these methods have different signature/results that need to be independently applied by the spatial map

jumanji/environments/swarms/search_and_rescue/env.py
zombie-einstein: Apart from the key passed to the target, all the functionality is deterministic, so all these keys are just dummy values that don't really need splitting. But yes you are correct, I should split at the top. I'll add a note to make this clear.

jumanji/environments/swarms/search_and_rescue/env.py
zombie-einstein: Can you elaborate, not sure I follow?

jumanji/environments/swarms/search_and_rescue/env.py
zombie-einstein: I think this need to be done twice (or at least I could not really see a way round) given there is a kind of asymmetry given a searcher can detect multiple targets inside one step (at least as it's implemented), and so we can't know the size of the values that would need storing ahead of time (for example we might try to store the indices of the searcher linked to each target to create rewards).

One workaround could be to only allow one detection, or have some fixed number of detections each step?

zombie-einstein: Very happy if you had another solution though, spent a while thinking about this 😆, this is one of the pain-points with JAX and simulations like this where interactions can variable size.

sash-a:
> Yeah could be an optional thing, or could just include by default and user can omit? Only issue might be impact of additional computation that may be unused?

Ye not sure, let's leave it for now and we can come back to it later if we feel it makes the problem too easy or is unnecessary

> I just added this functionality to Esquilax, and pulled changes into this so user can control the size of the space.

That's awesome! :fire: 

> Do you mind if resolve some of these comments. Was not sure if you wanted to use them for tracking, but it'd be handy if could clear up outdated or implemented ones to see what remains outstanding.

Ye go for it, if it's done then please resolve :smile:

docs/environments/search_and_rescue.md
zombie-einstein: Switched this over to set default as -1

jumanji/environments/swarms/search_and_rescue/env_test.py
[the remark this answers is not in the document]
sash-a: Ah I see ok then we can leave it at 22, hopefully it doesn't take too long :smile:

jumanji/environments/swarms/search_and_rescue/env.py
sash-a: The `shape` argument to `termination` and `transition` will set the shape of the discounts, but I realize this isn't necessary as agents will all finish at the same time so no need for separate discounts

jumanji/environments/swarms/search_and_rescue/env.py
sash-a: Ye it can be quite tricky, usually the solution is to use a mask. 

So in this case the mask would be `targets_found & ~targets.found[:, jnp.newaxis]` (I'm not 100% sure of if the newaxis should be dim 0 or 1) I think this should output a matrix of `(num_searchers, num_targets)` which would indicate which targets are found by which searchers in the current step. Then your reward for each searcher is just the sum along it's dimension.

jumanji/environments/swarms/search_and_rescue/dynamics.py
sash-a: Ye I think this would be easiest to keep the motion _smooth_

jumanji/environments/swarms/search_and_rescue/env.py
```diff
@@ -167,9 +170,12 @@ def __init__(
             view_angle=searcher_view_angle,
         )
         self.max_steps = max_steps
-        self._viewer = viewer or SearchAndRescueViewer()
         self._target_dynamics = target_dynamics or RandomWalk(0.01)
         self.generator = generator or RandomGenerator(num_targets=20, num_searchers=10)
+        self._viewer = viewer or SearchAndRescueViewer()
+        # Needed to set environment boundaries for plots
+        if isinstance(self._viewer, SearchAndRescueViewer):
+            self._viewer.env_size = (self.generator.env_size, self.generator.env_size)
```
sash-a, line 178: Instead of this could we pass the env size into `render`/`animate`

jumanji/environments/swarms/search_and_rescue/env.py
zombie-einstein: Oh yeah this is an option, my concern with this approach though is the need to carry flags for each target through the map reduce for each agent, so in total I think it would scale like `9 * n_searchers * n_targets` (the nine comes from the number of regions checked around each agent in parallel). For larger numbers this could cause some memory issues.

zombie-einstein: Though maybe mitigated if JAX has proper bit arrays?

jumanji/environments/swarms/search_and_rescue/env.py
zombie-einstein: Yeah, was not sure the best approach, thought it might be best to not have it set in 2 places, but happy to switch out.

jumanji/environments/swarms/search_and_rescue/env_test.py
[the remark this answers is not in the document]
zombie-einstein: Thinking about it, might make more sense to maybe shrink the space and add more agents. Mainly here just wanted to catch things like wrapping the agent co-ordinates and clipping there speeds velocities was working.

jumanji/environments/swarms/search_and_rescue/env.py
zombie-einstein: This may be overthinking/overoptimising though 😂 in scales of 10-100s of searchers to 100-1000s of targets this is in the Mb range (rather than Gb)

jumanji/environments/swarms/search_and_rescue/env.py
sash-a: Ye this just seems a bit brittle to me, eg what happens if we use another viewer for some reason.

jumanji/environments/swarms/search_and_rescue/env.py
sash-a: I'm confused though how does the current implementation scale differently? (This is likely due to my lack of understanding around spatial, I'm picturing a nested `vmap`?)

jumanji/environments/swarms/search_and_rescue/env_test.py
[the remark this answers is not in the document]
sash-a: But then more agents will also take longer, I think we can just leave it at 22, it's not *too* bad

jumanji/environments/swarms/search_and_rescue/env.py
zombie-einstein: Yeah it is effectively, so the space is subdivided into a grid (dependent on interaction range), then each agent gathers results from each grid cell, then those are reduced to a result for a single agent. So just adds an extra _fixed_ factor for results for each cell, so not really different from what you are saying.
Was thinking about the memory v computation trade-off, but in practise it's probably not that bad to just use the extra memory (for some reason had the memory usage being worse from a previous project)!

sash-a:
Hi @zombie-einstein, so I got it piping through in Mava unfortunately it doesn't seem to be learning there either, although I didn't leave it training for long and it was only feed forward independent PPO. I just did an env with 2 agents and 2 targets, maybe that is too easy? This is the branch: [feat/search-and-rescue](https://github.com/instadeepai/Mava/tree/feat/search-and-rescue) and [this](https://github.com/instadeepai/Mava/commit/71a768074a3a8545c263a27cecef293770f2dbc7) is the commit where you can see what I've changed.

Most of the important changes live in [`jumanji.py`](https://github.com/instadeepai/Mava/blob/71a768074a3a8545c263a27cecef293770f2dbc7/mava/wrappers/jumanji.py) and is where we wrap the env. An important thing to note mava expects all observations to be in the `agents_view` attribute of the `Observation` and the MLP expects flat obs (except for the agent and batch dim). Do note that I left out the time limit and number of targets remaining from the observation it is likely important to concat that onto the `agents_view`. You can change the config of the env in [this file](https://github.com/instadeepai/Mava/blob/71a768074a3a8545c263a27cecef293770f2dbc7/mava/configs/env/scenario/sar-default.yaml) where `task_config` will be passed to the generator or [here](https://github.com/instadeepai/Mava/blob/71a768074a3a8545c263a27cecef293770f2dbc7/mava/configs/env/search-and-rescue.yaml) where the `kwargs` will be passed directly to the env.

One thing I needed to do to get it working is to change the shape of the reward during reset - it doesn't contain the agent dimension during reset but it does during the step so I added it during reset.

```
        timestep = restart(observation=self._state_to_observation(state), shape=self.generator.num_searchers)
        timestep = timestep.replace(discount=timestep.discount[0])
```

Definitely not the cleanest though, maybe just repeat the reward instead of flattening the discount. 

If you could try get this training well either with Mava (use any algorithm you like) or with jumanji, I think that is the last blocker to merging this. If you have any questions about Mava feel free to ask here or open an issue there :smile: 


Here's the results I got after 1M steps
image

zombie-einstein:
Great thanks @sash-a, I'll grab a look, I suspect that as it stands with only 2 agents  this may be really difficult stochastic, the agents would see nothing until they bump into a target pretty much by chance. I actually just added an additional observation type that includes found and unfound targets in the view so I will try this first, should be easier (the different observations now treat this as if they have different visual channels for the different agent/target information).

sash-a:
Sounds great, I'd try other settings that make it easier also like increasing sight range and radius and generally anything else you can think of that would make it easier

zombie-einstein:
@sash-a I think latest JAX release 0.4.36 is breaking something, getting a bunch of `IndexError: list index out of range` in the CI. Testing locally 0.4.35 was running fine, but upgrade to 0.4.36 seems to fail. Want me to create a new PR to pin it less then 0.4.36?

sash-a:
Yes please! I was getting this morning in Mava also :smile: 

Seems to be related to https://github.com/jax-ml/jax/issues/25332

zombie-einstein:
Just wanted to track a couple final design choices I think have got a bit lost in the comments:

- Do we want to include some reward decay, to reward faster finding of targets
- Related to this do we want to include time/step in the observation?
- Do we want to add a velocity field to target states, not a massive change, and it will make it easier to extend target dynamics to something more complex

sash-a:
* I think having this an a different type of reward would be a good thing
* The convention is usually to add `step_count` to the observation and then you can also pass `time_limit` to the network so it can infer the remaining steps. On that note could you please change `max_steps` to `time_limit` as I'm trying to make this a convention.
* I think that would be nice, especially if you are finding that targets movement is unnatural or possibly hard to pursue. At least having it there in case we want to improve it in the future would be great!

zombie-einstein:
Hey @sash-a, it seems like at least for the fully visibility observation (i.e. agent can see unfound targets) this is relatively straightforward to train.
Below is for a single agent (pink) and 2 agents (blue) w 100 targets (that are mostly stationary)

image

single agent seems to be doing pretty well to get most of the targets, and w two agents they finish before the time limit. I'm guessing rewards recorded here are individual agents right? Only thing I want to double check is that for 2 agents rewards seem to max out at 50. This makes sense in that the total the agents can receive 100 in total, but it seems unlikely that one agent would randomly locate more targets than the other.

Will run more experiments, wanted to ask:

- Is there a way to pass arguments that are classes to the environment constructor in Mava? Can you make a wrapped constructor somewhere that can be passed arguments from a config?
- Is there a built in way to generate an animation during/after training?

sash-a:
> I'm guessing rewards recorded here are individual agents right?

The rewards are the mean over all agents, and summed over the episode.

> Is there a way to pass arguments that are classes to the environment constructor in Mava? Can you make a wrapped constructor somewhere that can be passed arguments from a config?

You can do this through `hydra.instantiate` and some config trickery, but it's much easier to  find the `environment.make` call in the system file and replacing it with exactly what you want, just be sure to add all the necessary wrappers in the order we do it in `utils.make_env`. We designed Mava to be hacked around with so I'd recommend this instead of changing the config.

> Is there a built in way to generate an animation during/after training?

No, but we have some scripts for this:

```
def _rollout_mava_system(
    env: Environment, params: FrozenDict, actor_state: ActorState, key: PRNGKey, act_fn: EvalActFn
) -> list:
    env_step = jax.jit(env.step)
    jit_act = jax.jit(act_fn)

    key, env_key = jax.random.split(key)
    state, ts = env.reset(env_key)

    states = []

    # Only loop once
    while not ts.last():
        # Eval env is wrapped in the record metrics wrapper. We just store the true env state
        # for jumanji to be able to render.
        states.append(state.env_state)
        ts = jax.tree_map(lambda x: x[jnp.newaxis, ...], ts)
        key, act_key = jax.random.split(key, 2)
        action, actor_state = jit_act(params, ts, act_key, actor_state)

        # note: dangerous squeeze, but we don't want a batch or time dim here
        state, ts = env_step(state, action.squeeze())

    return states
```

Then it's use in ff_ippo:

```
states = _rollout_mava_system(eval_env, tree.map(lambda x: x[0], trained_params), {}, key, eval_act_fn)
env.animate(states, save_path="your-path-here")
```

sash-a:
Hey just tracking that issue with JAX, it's been fixed in 0.4.37, so can you please unpin the JAX version in this PR

jumanji/environments/swarms/search_and_rescue/env.py
```diff
@@ -0,0 +1,381 @@
...
+        self._target_dynamics = target_dynamics or RandomWalk(0.001)
+        self.generator = generator or RandomGenerator(num_targets=100, num_searchers=2)
+        self._viewer = viewer or SearchAndRescueViewer()
+        self._reward_fn = reward_fn or SharedRewardFn()
+        self._observation = observation or AgentAndTargetObservationFn(
+            num_vision=64,
+            vision_range=0.1,
+            view_angle=searcher_view_angle,
+            agent_radius=0.01,
+            env_size=self.generator.env_size,
```
zombie-einstein, line 169: In this case do we want to pass these arguments in from the constructor arguments to make tweaking these values a bit a more streamlined the arguments are standardised across the different vision models? Though appreciate this is a standard pattern.

I did this when testing with Mava, passing in the type and constructing here.

jumanji/__init__.py
```diff
@@ -142,3 +142,10 @@
 # LevelBasedForaging  with a random generator with 8 grid size,
 # 2 agents and 2 food items and the maximum agent's level is 2.
 register(id="LevelBasedForaging-v0", entry_point="jumanji.environments:LevelBasedForaging")
+
+###
+# Swarm Environments
+###
+
+# Search-and-Rescue environment
+register(id="SearchAndRescue-v0", entry_point="jumanji.environments:SearchAndRescue")
```
zombie-einstein, line 151: Is it worth registering the environment with different vision models?

zombie-einstein:
Hi @sash-a, I just pushed a bunch of changes to add those features from the comments above (scaled rewards, time-step, and target velocities) so think this no mostly feature complete. Added a couple last comments above I wanted to run by you.

I also now want to do  a final pass overt the docstrings and docs, and add a couple more tests.

From testing, the environment with full visibility is relatively straightforward, but from the animation the agents seem to be essentially routing to the closest next unfound target (I need to find a way to upload a gif here!). 

I've not made much more progress with the model with hidden targets, though I need to test a few more configurations, I had a couple thoughts:

- The observation has no information about absolute agent locations, so it could be hard to learn an effective search pattern that attempts to cover the spae.
- It could be worth trying a global state that contains more information about agent location on the search area? Though his may be partially missing the point of the environment.

zombie-einstein:
This is with 4 searching agents, 100 targets and targets hidden until found. They seem to do pretty well and continue (slowly learning), though looking at their behaviours, again there's not a log of co-ordination, though this is probably to be expected from fully independent agents

image

jumanji/environments/swarms/search_and_rescue/env.py
sash-a: Ye it can be quite a mission to instantiate, it's possible we should reconsider this pattern, but for now leave it as is just to stay consistent

jumanji/__init__.py
sash-a: Do you mean fully versus partially observable?

sash-a: Ah I think you mean the different observation functions. I'd say we should aim to use the observation that only visualizes targets and searchers within their vision cones. If we see a good training curve with this then that should be the default.

In general I'd like to have a set of scenarios for most/all environments in jumanji (see #248). So it would be cool to think of a set (3-4) easy/hard envs and we can register those. If they happen to have different observation models that's fine

sash-a:
Hey @zombie-einstein this is really great progress and agreed it seems pretty much feature complete to me :fire:! Just a heads up I will be on holiday from the 16th of December to 6th January. I won't be able to do code review in that time, but I'll be available for any questions you may have. I think it's realistic to expect this to be merged early January. I'll just have a final look once I'm back and hopefully everything will be good to go :smile: 

> The observation has no information about absolute agent locations, so it could be hard to learn an effective search pattern that attempts to cover the spae.

Ok this is definitely an issue when learning and something I've noticed as being very important to the performance in other environments. I would add the absolute position of the current agent to its observation, and normalize it by world size.

> It could be worth trying a global state that contains more information about agent location on the search area? Though his may be partially missing the point of the environment.

Ye maybe the global state could be the absolute position of all agents and the target part of their observation?

> though this is probably to be expected from fully independent agents

I would highly recommend trying rec-mappo in mava (assuming there's a sensible global state) because this problem seems like it would definitely benefit from some recurrence in the policy.

jumanji/__init__.py
zombie-einstein: Yeah I was picturing something along these lines, like the easy one is where un-found targets are visible, and then harder versions use the version with hidden targets.

zombie-einstein:
> Ok this is definitely an issue when learning and something I've noticed as being very important to the performance in other environments. I would add the absolute position of the current agent to its observation, and normalize it by world size.

> I would highly recommend trying rec-mappo in mava (assuming there's a sensible global state) because this problem seems like it would definitely benefit from some recurrence in the policy.

Great yeah this makes sense, will add this and try it out. 

Just to check, the global state should have one entry per agent, something like `[n-agents, global-obs]`? Thinking if the position part of the observation needs to be specific to each agent, for example if the positions are an array

```
[p0, p1, p2, ....]
```
the per-agent position observation might need to be rotated like?
```
[
    [p0, p1, p2, ...., pn],
    [p1, p2,  ..., pn, p0],
    [p2, ...., pn, p0, p1],
]
```

sash-a:
> Just to check, the global state should have one entry per agent, something like [n-agents, global-obs]? Thinking if the position part of the observation needs to be specific to each agent, for example if the positions are an array

That's the way we do it in mava, 1 per agent. But to be honest it's a bit of an open question as to how the global state should be structured. In my opinion it should be the same across all agents (I'm pretty sure this is what the theory says) but empirically we've seen good results when it's tailored to each agent. So what I would do for a first pass is a global state of shape `(num_agents, global_state shape)` where it is the same along the first axis if this doesn't work well then try tailor it per agent.

Also I'm assuming this was just an example above, but don't forget the global state should include information about targets. Some global state ideas for targets, in all cases these include the normalised absolute positions of all agents:

* Absolute positions of all targets if they are visible to _any_ agents otherwise some out of bounds default value 
* Always adding target positions to the global state regardless of if they are visible (this one is probably cheating a bit though). 
* The target part of all agents observations

Honestly often concatenating all observations is a good global state also, it's hard to know what will work best without testing it.

sash-a:
Hi @zombie-einstein hope you had a great new year! Just wondering what the status of this is, how is the training looking?

zombie-einstein:
> Hi @zombie-einstein hope you had a great new year! Just wondering what the status of this is, how is the training looking?

Happy new year @sash-a, hope you've had a good break. I've been catching up with some other work over the break (including some tweaks to Esquilax like not requiring random keys), I got the training working but need to tweak the shared observation. It's on my list to hopefully finalise this week.

zombie-einstein:
Hey @sash-a, so some progress with this with some sensible looking training curves (though I think there is like some improvement to be made). 

So this is with 

- 4 agents
- 25 targets
- Rewards scaled (linearly decreasing) by time
- Recursive MAPPO
- Agents share their view on targets and global agent positions
- Greedy evaluation

image

You can see the agents completing the search more efficiently over the course of training. 

Without the reward scaling, I've been struggling to get similar results, possibly the task is still kind of stochastic for individual agents, without some more specific search planning/scheduling.

sash-a:
@zombie-einstein these curves are looking great! 

So when you say linearly scaled rewards, do you mean that finding a target early gives more reward than finding one later on? If so I'd say that's perfect.

It does seem to peak quite early, which may just be that the current env configuration is a bit easy? Can you try tweaking the parameters to make it a bit harder, maybe more agents or a smaller fov? 

How many steps was this run for, was it the mava default?

Given these curves I am pretty much happy to merge, I will give it another once over next week and hopefully will be able to merge after that!

zombie-einstein:
> @zombie-einstein these curves are looking great!
> 
> So when you say linearly scaled rewards, do you mean that finding a target early gives more reward than finding one later on? If so I'd say that's perfect.

Yeah exactly this, though this is configurable and allows custom implementations of the scaling.

> It does seem to peak quite early, which may just be that the current env configuration is a bit easy? Can you try tweaking the parameters to make it a bit harder, maybe more agents or a smaller fov?

Yeah I think there is kind of simple local optima which is to move to explore the space (i.e. not just sit in one spot or circle) and then bump into the targets. Feels like a really consistent strategy would be pretty involved, i.e. having a way of truly covering the space. Sure I'll try a couple more permutations, could reduce the number of targets again, or reduce agent speed.

> How many steps was this run for, was it the mava default?

I increased `num_updates` to 2,000, but think I need to also play with some of the other parameters? I also increased the critic parameters to handle the larger global observation.

> Given these curves I am pretty much happy to merge, I will give it another once over next week and hopefully will be able to merge after that!

Nice one, I'll try grab a look over it before since there's been a lot of changes.

sash-a:
> Yeah I think there is kind of simple local optima which is to move to explore the space (i.e. not just sit in one spot or circle) and then bump into the targets. Feels like a really consistent strategy would be pretty involved, i.e. having a way of truly covering the space. Sure I'll try a couple more permutations, could reduce the number of targets again, or reduce agent speed.

Ye makes sense, it's definitely a hard problem. Great edit whichever parameters you feel makes the most sense

> I increased num_updates to 2,000, but think I need to also play with some of the other parameters? I also increased the critic parameters to handle the larger global observation.

Ah interesting, then that's not such a surprising plateau as that would give you around 40m timesteps 

> Nice one, I'll try grab a look over it before since there's been a lot of changes.

Great, let me know when you're ready for me to have a look again :smile:

zombie-einstein:
Hi @sash-a, think this is ready to look over now. Only remaining task is to add an animated image to the docs. 

There's also a mypy bug come up due to recent [change in MyPy](https://mypy.readthedocs.io/en/stable/changelog.html#change-to-enum-membership-semantics) that was causing CI to fail so I grabbed that here.

jumanji/environments/swarms/search_and_rescue/env.py
```diff
@@ -0,0 +1,395 @@
...
+    boundaries.
+
+    An episode will terminate if all targets have been located by the team of
+    searching agents.
+
+    - observation: `Observation`
+        searcher_views: jax array (float) of shape (num_searchers, channels, num_vision)
+            Individual local views of positions of other agents and targets, where
+            channels can be used to differentiate between agents and targets.
+            Each entry in the view indicates the distant to another agent/target
```
sash-a, line 61: grammar
```suggestion
            Each entry in the view indicates the distance to another agent/target
```

jumanji/environments/swarms/search_and_rescue/env.py
```diff
@@ -0,0 +1,395 @@
...
+            along a ray from the agent, and is -1.0 if nothing is in range along the ray.
+            The view model can be customised using an `ObservationFn` implementation, e.g.
+            the view can include agents and all targets, agents and found targets,or
```
sash-a, line 64: grammar
```suggestion
            the view can include agents and all targets, agents and found targets, or
```

jumanji/environments/swarms/search_and_rescue/env.py
```diff
@@ -0,0 +1,395 @@
...
+        accelerate/decelerate the agent as [rotation, acceleration] on the range
+        [-1, 1]. These values are then scaled to update agent velocities within
+        given parameters (i.e. a value of -+1 is the maximum acceleration/rotation).
+
+    - reward: jax array (float) of shape (num_searchers,)
+        Arrays of individual agent rewards. A reward of +1 is granted when an agent
+        comes into contact range with a target that has not yet been found, and
+        the target is within the searchers view cone. Rewards can be shared
+        between agents if a target is simultaneously detected by multiple agents,
+        or each can be provided the full reward individually.
```
sash-a, line 83: Can you add a similar sentence here about customizing the reward with `RewardFn`s same as you did for the observation

jumanji/environments/swarms/search_and_rescue/env.py
```diff
@@ -0,0 +1,395 @@
...
+                a value from [0,1] representing a fraction of π-radians.
+                The view cone of an agent goes from +- of the view angle
+                relative to its heading, e.g. 0.5 would mean searchers have a
+                90° view angle in total.
+            time_limit: Maximum number of environment steps allowed for search.
+            viewer: `Viewer` used for rendering. Defaults to `SearchAndRescueViewer`.
+            target_dynamics: Target object dynamics model, implemented as a
+                `TargetDynamics` interface. Defaults to `RandomWalk`.
+            generator: Initial state `Generator` instance. Defaults to `RandomGenerator`
+                with 20 targets and 10 searchers.
```
sash-a, line 149: Note set this default to what ever you found worked best in your tests (same for all defaults)

jumanji/environments/swarms/search_and_rescue/env.py
```diff
@@ -0,0 +1,395 @@
...
+
+        Returns:
+            state: Updated searcher and target positions and velocities.
+            timestep: Transition timestep with individual agent local observations.
+        """
+        key, target_key = jax.random.split(state.key, num=2)
+        searchers = update_state(
+            self.generator.env_size, self.searcher_params, state.searchers, actions
+        )
+        targets = self._target_dynamics(target_key, state.targets, self.generator.env_size)
```
sash-a, line 236: I wonder if this ordering isn't a bit unfair. I think there's a possible scenario where an agent moves towards a target, and then that target moves in the opposite direction - behind the agent where it can no longer see it. In this case would the agent be rewarded. Maybe a better ordering is: take actions -> resolve if found -> get rewards -> move targets. Although this is a significant change, what do you think?

jumanji/environments/swarms/search_and_rescue/env.py
```diff
@@ -0,0 +1,395 @@
...
+            self.searcher_params.view_angle,
+            searchers,
+            (jnp.arange(self.generator.num_targets), targets),
+            pos=searchers.pos,
+            pos_b=targets.pos,
+            env_size=self.generator.env_size,
+            n_targets=self.generator.num_targets,
+        )
+
+        rewards = self._reward_fn(targets_found, state.step, self.time_limit)
```
sash-a, line 255: Maybe I'm missing something, doesn't the reward function expect targets found this step? These targets could have also been found on previous steps. `targets_found & ~state.targets.found[jnp.newaxis]` makes more sense to me.

jumanji/environments/swarms/search_and_rescue/env.py
```diff
@@ -0,0 +1,395 @@
...
+            min_speed=searcher_min_speed,
+            max_speed=searcher_max_speed,
+            view_angle=searcher_view_angle,
+        )
+        self.time_limit = time_limit
+        self._target_dynamics = target_dynamics or RandomWalk(0.001)
+        self.generator = generator or RandomGenerator(num_targets=50, num_searchers=2)
+        self._viewer = viewer or SearchAndRescueViewer()
+        self._reward_fn = reward_fn or SharedRewardFn()
+        self._observation = observation or AgentAndAllTargetObservationFn(
```
sash-a, line 171: Can we call this `_observer` or `_observation_fn`

sash-a (commented):
I haven't looked at the training stuff, does it work or did you end up using mava? If so I think you can remove the training stuff and keep only the random agent

Otherwise just a few questions where I think there might be mistakes. Once this is done and the gif is added I want to try run it locally, what branch do you have it working on mava?

zombie-einstein:
> I haven't looked at the training stuff, does it work or did you end up using mava? If so I think you can remove the training stuff and keep only the random agent

The training stuff in this PR works well where all the targets are visible, but I ended up using Mava for the harder problem where targets are only shown once found (so I could use a recursive network and the centralised observation).

> Otherwise just a few questions where I think there might be mistakes. Once this is done and the gif is added I want to try run it locally, what branch do you have it working on mava?

I was working off of the branch [you started](https://github.com/instadeepai/Mava/tree/feat/search-and-rescue), but I actually meant to ask if you can add me as a contributor to that branch, then I can push changes I've made.

jumanji/environments/swarms/common/updates.py
```diff
@@ -0,0 +1,207 @@
...
+    """
+    actions = jnp.clip(actions, min=-1.0, max=1.0)
+    headings, speeds = update_velocity(params, (actions, state))
+    positions = jax.vmap(move, in_axes=(0, 0, 0, None))(state.pos, headings, speeds, env_size)
+
+    return types.AgentState(
+        pos=positions,
+        speed=speeds,
+        heading=headings,
+    )
```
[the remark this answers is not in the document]
zombie-einstein: Yup good shout

jumanji/environments/swarms/search_and_rescue/observations.py
```diff
@@ -0,0 +1,392 @@
...
+        vision_range: float,
+        view_angle: float,
+        agent_radius: float,
+        env_size: float,
+    ) -> None:
+        """
+        Vision model that contains other agents and found targets.
+
+        Searchers and targets are visualised as individual channels.
+        Targets are only included if they have been located already.
```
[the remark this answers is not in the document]
zombie-einstein: Not sure I follow. This observation only shows located targets and then there is also the `AgentAndAllTargetObservationFn` that show both targets found and not (the easier task).

jumanji/environments/swarms/search_and_rescue/env.py
zombie-einstein: Thanks good spot

jumanji/environments/swarms/search_and_rescue/env.py
zombie-einstein: I think this is correct. So rewards in this line is newly found targets in this step, the map over the searchers checks the current status and only sets a flag when it's newly found. Then the state is updated in line `259`. This [test](https://github.com/zombie-einstein/jumanji/blob/ac3f811e023108e9c3a538fc72773dcd32f5910c/jumanji/environments/swarms/search_and_rescue/env_test.py#L164) should catch this error?

jumanji/environments/swarms/search_and_rescue/dynamics.py
```diff
@@ -0,0 +1,69 @@
...
+
+        Args:
+            key: Random key.
+            targets: Current target states.
+            env_size: Environment size.
+
+        Returns:
+            Updated target states.
+        """
+        d_pos = jax.random.uniform(key, targets.pos.shape)
```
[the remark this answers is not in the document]
zombie-einstein: For basic usage and small I values I think this is ok, though can see that this could be jumpy (and exacerbate some of the update orderings you've highlighted below).

Did you have another model in mind? Happy to add some more options.

jumanji/environments/swarms/search_and_rescue/dynamics.py
```diff
@@ -0,0 +1,69 @@
...
+        d_pos = self.step_size * 2.0 * (d_pos - 0.5)
+        pos = (targets.pos + d_pos) % env_size
+        return TargetState(pos=pos, vel=targets.vel, found=targets.found)
```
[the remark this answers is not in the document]
zombie-einstein: Ah yeah I think the reason I avoided this is that there's a discrepency between the actual chex dataclass that used `replace`, and the python dataclass that does not when type checking. I've change with the annotation to skip typing.

jumanji/environments/swarms/search_and_rescue/env.py
zombie-einstein: Happy to change though, don't think this is too massive a change

sash-a:
> The training stuff in this PR works well where all the targets are visible, but I ended up using Mava for the harder problem where targets are only shown once found (so I could use a recursive network and the centralised observation).

Ok great, I'll review that later this week then.

> I was working off of the branch [you started](https://github.com/instadeepai/Mava/tree/feat/search-and-rescue), but I actually meant to ask if you can add me as a contributor to that branch, then I can push changes I've made.

Unfortunately I can't add you as a contributor to Mava, but if you fork the repo you should be able to commit your changes to the forked repo

jumanji/environments/swarms/search_and_rescue/dynamics.py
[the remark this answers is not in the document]
sash-a: My only other suggestion is a similar model with a change in velocity - so sample a normal distribution centered around the current velocity and then clamp the output within a max and min.

However I don't think it's super necessary to add this

jumanji/environments/swarms/search_and_rescue/observations.py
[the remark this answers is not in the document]
sash-a: I see, shouldn't targets always be visible if they are within the field of view and close enough, otherwise the agents will need to randomly bump into targets to find them. This way they can look around to locate a target. 

You've tested this though, so does this make it too easy? Maybe doing it this way + having a smaller fov and many targets/searchers would make it hard enough?

jumanji/environments/swarms/search_and_rescue/env.py
sash-a: Ye it might cause a disconnect and I feel my scenario would almost never happen, so happy to keep it as is

jumanji/environments/swarms/search_and_rescue/env.py
sash-a: Ah I see, I missed this `~target.found` in `searcher_detect_targets`

jumanji/environments/swarms/search_and_rescue/observations.py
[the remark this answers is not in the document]
zombie-einstein: Yeah I see your point. When all the targets are visible it does seem kind of easy, it becomes more like a routing task where the agents just move to the next closest target (not that this could not be useful task itself) but yeah maybe a smaller fov is a good idea. With hidden targets it seems more like a task to efficiently cover the space (though from testing yes this is much more difficult).

One thing we could consider is that at the moment the vision-range of the searchers is the same irrespective of what they are viewing. It felt sensible to have the searchers have a decent vision range of other searchers (if they are working as a team), but then this mean the vision range for targets on the same scale makes locating targets relatively easy. Could break these into two value so there's finer control over the view range. We'd then have 3 parameters

- Searcher vision range
- Target vision range
- Target interaction range

With generally searcher-range > target-range. So then targets would be visible if within the target vision, but not marked found until in the interaction range?

jumanji/environments/swarms/search_and_rescue/reward.py
```diff
@@ -0,0 +1,109 @@
...
+    rewards = jnp.where(norms > 0, rewards / norms, rewards)
+    return rewards
+
+
+def _scale_rewards(rewards: chex.Array, step: int, time_limit: int) -> chex.Array:
+    scale = (time_limit - step) / time_limit
+    return scale * rewards
+
+
+class SharedRewardFn(RewardFn):
```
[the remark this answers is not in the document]
sash-a: Honestly I would just keep 1 of the two types of rewards (which ever you've found to work best) and then have a scaled and non-scaled version.

zombie-einstein:
Hey @sash-a merged fixes for above comments. Consolidated a lot of the observation and reward functionality, and added in distinct target and agent vision ranges. Need to maybe tweak some of the default parameters but otherwise should be ready to look over.

zombie-einstein:
> All these changes look great! Question is how does the learning perform now?

image

So this is 4 searchers and 40 targets, with a narrower FOV and the smaller target vision range, using `rec_ippo` in Mava. So they continue to get quicker at completing the task, though the relative increase in the rewards slows. 

There's maybe some tweaking to do with the parameters, performance seems to be relatively sensitive to small changes. 

With the larger observation space I was actually struggling to get the shared observation for Mappo working effectively.

I'll make sure to push my Mava changes to fork today!

zombie-einstein:
@sash-a my Mava fork is [here](https://github.com/zombie-einstein/Mava)

zombie-einstein:
rec-mappo (green) working now, but not gain over rec-ippo (yellow)

image

sash-a:
Thanks @zombie-einstein this looks amazing. Maybe a little easy given how quickly it seems to find the optimal strategy, but it clearly is learning! Honestly having an easy default isn't a bad thing. 

I'm pretty much happy with all the code, I just want to try run your fork locally and play around with the difficulty myself. My week is just a bit busy, but I will get to it next week and then we should be able to merge it! :rocket: 

(Sorry this has taken so long)

docs/environments/search_and_rescue.md
```diff
@@ -0,0 +1,81 @@
+# 🚁 Search & Rescue
+
+[//]: # (TODO: Add animated plot)
```
sash-a, line 3: Can you please add the gifs and remove this TODO

sash-a (commented):
Just some minor changes to the distributions and also could you please add the gif and image and update the readme :pray:

jumanji/training/networks/postprocessor.py
```diff
@@ -47,6 +48,23 @@ def forward_log_det_jacobian(self, x: chex.Array) -> chex.Array:
...
+    def forward(self, x: chex.Array) -> chex.Array:
+        return self._tanh.forward(x)
+
+    def inverse(self, y: chex.Array) -> chex.Array:
+        return self._tanh.inverse(y)
+
+    def forward_log_det_jacobian(self, x: chex.Array) -> chex.Array:
+        return self._tanh.forward_log_det_jacobian(x)
+
+
```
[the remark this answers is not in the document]
zombie-einstein: Oh yeah, thought this might cause type check issues, but will change if not!

jumanji/training/configs/env/search_and_rescue.yaml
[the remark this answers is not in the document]
zombie-einstein: Sure will do

docs/environments/search_and_rescue.md
zombie-einstein: Sure will get on it!

zombie-einstein:
Hi @sash-a, think that's all the comments addressed in the last merge

sash-a:
Sorry last bit of admin, I'll do this as it seems I have access to your repo:
* Readme table needs to be updated with the search and rescue env
* We need an env image in docs/env_img
* That image needs to be added to the existing images at the top of the readme


PS: I see the targets were changing colour when they were found already, it must have just not been happening in the random episode I visualized - I prefer the way you had it originally, I'll change it back :joy:

zombie-einstein:
> PS: I see the targets were changing colour when they were found already, it must have just not been happening in the random episode I visualized - I prefer the way you had it originally, I'll change it back 

What was the issue here? I used the updated version to produce the animation, so thought it was working ok?

zombie-einstein (commented):
Thanks, just realised I only testing the training with random policy after this change!

sash-a:
> What was the issue here? I used the updated version to produce the animation, so thought it was working ok?

There was no issue, I didn't realize that it was changing colours before because I was visualizing a random episode that never found any targets. So I just reverted it to what you had originally and made the colours a little nicer

sash-a (approved):
Thanks for the amazing work on this @zombie-einstein it is very much appreciated! :1st_place_medal:

zombie-einstein:
> Thanks for the amazing work on this @zombie-einstein it is very much appreciated! 🥇

Awesome, thanks @sash-a 🎉, thanks for your time an patience also!

I've an idea for another environment that could use quite a few of the mechanics from this, so will open an issue soon(ish).

sash-a:
Awesome hopefully subsequent PRs will be quicker!

Issue #255, opened 2024-10-29
Switch to ruff for linting and formatting


sash-a:
Pretty much title. Not hard to do, just a lot of fixing required once we switch.

PR #260, opened 2024-11-04
feat: switch to ruff and upgrade pre-commit hooks

sash-a:
Switch from black/flake/isort to ruff and update pre-commit hooks

Closes: #255


WiemKhlifi (approved):
🧼 🚀

Issue #256, opened 2024-11-03
Add dtype choice in step type/functions


thomashirtz:
### Is your feature request related to a problem? Please describe
In a personal project, I had to be very efficient in the memory management, my reward were taking a lot of space. In my case I needed to change the reward and the discount to float16 instead of float32. I had to copy over the types file to do the modification locally. However I feel like some use case may need this flexibility.

### Describe the solution you'd like
Give an extra parameter to the step functions to give dtype. (Example with one of them)

```python
def truncation(
    reward: Array,
    observation: Observation,
    discount: Optional[Array] = None,
    extras: Optional[Dict] = None,
    shape: Union[int, Sequence[int]] = (),
    dtype: jnp.dtype = jnp.float32,
) -> TimeStep:
    """Returns a `TimeStep` with `step_type` set to `StepType.LAST`.

    Args:
        reward: array.
        observation: array or tree of arrays.
        discount: array.
        extras: environment metric(s) or information returned by the environment but
            not observed by the agent (hence not in the observation). For example, it
            could be whether an invalid action was taken. In most environments, extras
            is None.
        shape: optional parameter to specify the shape of the rewards and discounts.
            Allows multi-agent environment compatibility. Defaults to () for
            scalar reward and discount.
    Returns:
        TimeStep identified as the truncation of an episode.
    """
    discount = discount if discount is not None else jnp.ones(shape, dtype=dtype)
    extras = extras or {}
    return TimeStep(
        step_type=StepType.LAST,
        reward=reward,
        discount=discount,
        observation=observation,
        extras=extras,
    )
```

I would be happy to do the PR.

---
### Misc
- [x] Check for duplicate requests.

sash-a:
I think that would be a nice addition, happy to review it :smile:

sash-a:
To be honest I think the discounts should probably be booleans while they are stored in the timestep because for me they just indicated end of episode, but I think this would add nice flexibility

sash-a:
My only issue is that this strays from the original [dm_env api](https://github.com/google-deepmind/dm_env/blob/master/dm_env/_environment.py#L148-L157) where it is a float so it can represent both RL discount (gamma) and done. 

Let's definitely add it as an argument, but for the default I'm not sure if boolean or float32 is best @clement-bonnet any thoughts on this?

clement-bonnet:
[...]

PR #262, opened 2024-11-05
Add dtype choice in step type/functions

thomashirtz:
See #256 

I just feel that it is a little weird that this dtype argument is both responsible for both reward and discount in "restart". But otherwise it means that there are no way to choose reward's dtype at restart time.

### DIFF
--- a/jumanji/types.py
+++ b/jumanji/types.py
@@ -108,6 +108,8 @@ def restart(
         shape: optional parameter to specify the shape of the rewards and discounts.
             Allows multi-agent environment compatibility. Defaults to () for
             scalar reward and discount.
+        dtype: Optional parameter to specify the data type of the rewards and discounts.
+            Defaults to `float`.
 
     Returns:
         TimeStep identified as a reset.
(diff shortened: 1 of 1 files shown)


jumanji/types.py
```diff
@@ -95,6 +95,7 @@ def restart(
     observation: Observation,
     extras: Optional[Dict] = None,
     shape: Union[int, Sequence[int]] = (),
+    dtype: Union[jnp.dtype, type] = float,
```
sash-a, line 98: Can you please add this to the docs for all these methods :pray:

sash-a (changes requested):
Thanks for this @thomashirtz 

Can you please add the dtype arg to the docstring and also sign the CLA :smile:

jumanji/types.py
thomashirtz: Oh mb, I forgot about it, it's done, tell me if it is okay :)

sash-a (approved):
Thanks @thomashirtz. Can you please sign the contributor license agreement, click the link in the first comment :pray:

PR #266, opened 2024-11-25
fix: fix termination condition

WiemKhlifi:
# What:
Fix the termination condition in the `JumanjiToGymWrapper`.


sash-a (approved):
Good spot, thanks!

PR #269, opened 2024-12-09
fix: Require jax below 0.4.36

zombie-einstein:
Restrict JAX to below latest 0.4.36 release due to [this issue](https://github.com/jax-ml/jax/issues/25332)


sash-a (approved):
:fire:

Issue #253, opened 2024-10-29
Upgrade to latest matplotlib


sash-a:
Matplotlib's animation API has changed in new releases, we should try updating to the latest version

sash-a:
Done in #270

PR #270, opened 2024-12-13
Matplotlib upgrade

zombie-einstein:
Hey @sash-a, thought I'd grab a look at #253.

Changes here update Matplotlib version and then makes changes to pass type checking, so it builds and passes tests.

_But_ ...

The new API now requires that the animation update function returns a list of artists that have been updated, these are objects returned from plotting calls like a plot object, or text box (see the table [here](https://matplotlib.org/stable/users/explain/animations/animations.html#funcanimation)). The state/data of these objects can be updated without redrawing the whole thing.

I've done this rubiks cube environment:

```python
# Create the first frame of the plot and return a list of drawn image artists
images = self._draw(ax, states[0])

def make_frame(state: State) -> Sequence[Artist]:
    # Update the data in each image from the state
    for i, image in enumerate(images):
        image.set_data(state.cube[i])
    # Return the updated images
    return images

# Create the animation object.
self._animation = matplotlib.animation.FuncAnimation(
    fig,
    make_frame,
    frames=states,
    interval=interval,
)
```

Doing this for all environments is easier for some than others given how they've been written, it means passing all the relevant artists to the animate function when the plot is created, and then correctly updating them inside the update function.

I think the animations will still work as is, as they still generally update the axes by reference inside the animation function, it's just less efficient then the new API that just updates the relevant artists.


sash-a:
Wow this is an awesome addition, thanks @zombie-einstein!

Which environments are still remaining?

Also I think the way you've done it with `make_frame` looks good, but I have no experience with the new API

zombie-einstein:
> Which environments are still remaining?

Apart from rubix cube they all need a look at, with varying degrees of complexity. I think the general pattern needs to be something like a `draw` function that draws the first frame (and also draws the individual frames via the `render` method, and returns any actors that will be updated during the animation, then use this to initialise the first frame and update actors inside the update-func.

Some also currently have actors that are created dynamically during plotting I've been trying to work out the best pattern for, in some cases it seems it would be more efficient to initialise all the actors ahead of time, and update their states as needed. E.g. for the sudoko environment initialise all the text actors and update the text duting animation (rather than create them on the fly).

So there's a decent chunk of work to refactor _all_ the plots. One option could be merge something like this PR, which will updated the requirement and passes linting/tests, and still create plots (just less efficiently), and then incrementally update the individual viewers in separate PRs to make best use of the new API.

sash-a:
Hi @zombie-einstein sorry for leaving this hanging, I needed a bit of time off :joy: 

I'll have a deeper look at it this week and give some a proper opinion :smile:

zombie-einstein:
> Hi @zombie-einstein sorry for leaving this hanging, I needed a bit of time off 😂
> 
> I'll have a deeper look at it this week and give some a proper opinion 😄

Sure no worries, let me know if you've any follow up questions on the API update.

jumanji/environments/logic/rubiks_cube/viewer.py
```diff
@@ -70,15 +74,18 @@ def animate(
...
 
-        def make_frame(state_index: int) -> None:
-            state = states[state_index]
-            self._draw(ax, state)
+        images = self._draw(ax, states[0])
+
+        def make_frame(state: State) -> Sequence[Artist]:
+            for i, image in enumerate(images):
+                image.set_data(state.cube[i])
+            return images
```
sash-a, line 82: This was a bit confusing on first read, because I thought you somehow got all images from the `_draw` method, so I'd rather call it faces

```suggestion
        faces = self._draw(ax, states[0])

        def make_frame(state: State) -> Sequence[Artist]:
            for i, face in enumerate(faces):
                face.set_data(state.cube[i])
            return faces
```

jumanji/environments/logic/rubiks_cube/viewer.py
```diff
@@ -103,9 +110,10 @@ def _get_fig_ax(self) -> Tuple[plt.Figure, List[plt.Axes]]:
                 fig.show()
         return fig, ax
 
-    def _draw(self, ax: List[plt.Axes], state: State) -> None:
-        i = 0
-        for face in Face:
+    def _draw(self, ax: List[plt.Axes], state: State) -> List[AxesImage]:
+        images = list()
```
sash-a, line 114: nit, bit more pythonic
```suggestion
        images = []
```

jumanji/environments/packing/tetris/viewer.py
```diff
@@ -197,7 +198,7 @@ def animate(
         fig, ax = plt.subplots(num=f"{self._name}Animation", figsize=TetrisViewer.FIGURE_SIZE)
         plt.close(fig)
 
-        def make_frame(grid_index: int) -> None:
+        def make_frame(grid_index: int) -> Tuple[Artist]:
```
sash-a, line 201: hmmmm this goes against the pattern you've set up, but I can see there's some weird filtering going on in that for loop, so I'm not sure there's a good way around it?

requirements/requirements.txt
```diff
@@ -3,7 +3,7 @@ dm-env>=1.5
 gymnasium>=1.0
 huggingface-hub
 jax>=0.2.26,<0.4.36
-matplotlib~=3.7.4
+matplotlib>3.8.0
```
sash-a, line 6: Should it not be `>=` I'm not entirely sure here, but I think the animation update was in 3.8?

sash-a (commented):
I like the pattern, although seems like tetris is a special case?

One thing that I think would be nice to standardise is the name of the main draw function, many are called `_add_foo` or `_draw_xyz`. I think a good name would be `_draw_state` or `_draw_frame`?

As a function signature maybe:
```
def _draw_state(state: State, artist: Artist) -> Artist
```

Just going to check that everything is rendering properly

sash-a:
Ok I've run all of them and they look good to me except for:

* graph colouring
* knapsack 
* MMST
* sodoku
* TSP

Knapsack seems to just be because the knapsack image isn't added to the manifest which is an easy fix, the others seem to have deeper issues. I tested with the [following notebook](https://colab.research.google.com/drive/17sreU-e2e8mp5ML9HbIQVG8Vj9uNpNRM?usp=sharing) (we can add this to the examples folder for easy testing in the future)

zombie-einstein:
> One thing that I think would be nice to standardise is the name of the main draw function, many are called `_add_foo` or `_draw_xyz`. I think a good name would be `_draw_state` or `_draw_frame`?
> 
> As a function signature maybe:
> 
> ```
> def _draw_state(state: State, artist: Artist) -> Artist
> ```

Yeah think this would be ideal, I think it should be something like a 'draw_frame(...) -> Sequence[Artist]` that draws the full frame/image and can be called to draw a single frame/image and returns any artists that can be updated (I think in principle this could be any data structure of artists).

requirements/requirements.txt
zombie-einstein: Thanks, yeah I struggled a little to locate the exact release from the release notes

jumanji/environments/packing/tetris/viewer.py
zombie-einstein: Yeah, I guess there's some flexibility in the implementation so not like it needs to be a hard rule (I think), there's likely some way to consolidate the updates into a single iterable.

jumanji/environments/logic/rubiks_cube/viewer.py
zombie-einstein: Could have sworn that `list()` was preferred for some pep reason, but looking into think you are correct. Must have imagined it 😂!

zombie-einstein:
> Ok I've run all of them and they look good to me except for:
> 
>     * graph colouring
> 
>     * knapsack
> 
>     * MMST
> 
>     * sodoku
> 
>     * TSP
> 
> 
> Knapsack seems to just be because the knapsack image isn't added to the manifest which is an easy fix, the others seem to have deeper issues. I tested with the [following notebook](https://colab.research.google.com/drive/17sreU-e2e8mp5ML9HbIQVG8Vj9uNpNRM?usp=sharing) (we can add this to the examples folder for easy testing in the future)

Yeah good shout, it was kind of painful to check all these manually.

jumanji/environments/logic/graph_coloring/viewer.py
```diff
@@ -65,26 +57,42 @@ def animate(
...
-
-        def make_frame(state_index: int) -> None:
-            state = states[state_index]
-            self.render(state, ax=ax)
+        self._set_params(states[0])
+        fig = plt.figure(f"{self._name}Animation", figsize=self.FIGURE_SIZE)
+        plt.subplots_adjust(left=0.0, right=1.0, top=1.0, bottom=0.0)
+        ax = fig.add_subplot(111)
+        plt.close(fig)
+        circles = self._prepare_figure(ax, states[0])
```
zombie-einstein, line 65: So this seems to have repaired the graph-coloring viewer, just need to check a couple details:

- Graph edges are constant?
- Do we need to color the edges? Seemed like they were updated, but I was not sure the intention (this is relatively easily added)

zombie-einstein:
Hey @sash-a I fixed the graph-colouring, MMST, sodoku, and  TSP viewers (still need to work out the img path for the knapsack viewer). 

Refactoring the graph based viewers, it seems like the intended pattern is to create all the artists ahead of time and then update their properties. This raises a couple potential difficulties:

- Environments with variable numbers of displayed artists become somewhat of a pain, I guess the correct thing to do is to create the max number of artists you might need ahead of time, and then display them as needed.
- Is the animate function intended to display states over multiple episodes? For example for the MMST env, the structure of the graph is resampled when the environment is reset. It's nice if the edges were static, then you don't need to recalculate the node positions etc and can just update the edge & node properties, and also save on the relatively node position calculation. But then this causes an issue if the states span multiple episodes. There's maybe a way to do this where you check if the episode has terminated, though a bit more involved.

sash-a:
So it currently does work over multiple episodes and I think we should try keep that functionality. Maybe there is a way to tell if an episode is done from the state?

sash-a:
For knapsack what should work is if you follow the method used in [TSP for loading images](https://github.com/instadeepai/jumanji/blob/main/jumanji/environments/routing/tsp/viewer.py#L138-L140). I think all that should be needed to move the file into the jumanji repo instead of reading it from docs as the `manifest` automatically includes png files which the knapsack image is

We [recently fixed TSP](https://github.com/instadeepai/jumanji/pull/251) and I assume missed knapsack.

sash-a:
I would go with the lower impact option as the other option is a big API change. I don't think it's a big issue if the rendering is a bit slower than it could be :shrug:

zombie-einstein:
Hey @sash-a, I think this is all working now, though some of the code definitely needs a work over. Turned out to be a bit more involved:

- I went with checking for a new episode from changes in state, then updating the static elements for a new episode
- I worked out how to add and remove artists inside the update function, so used this in the cases where the number of artists changes, e.g. routes in tsp
- In some cases there was an issue where old elements of the plot were not be in removed after a new episode. I think I've addressed these and the plots now all have the expected behaviour.
- I've not fully updated all the animations, in the case where they seem to work using the old functionality I've just re-used that.

I also grabbed a couple other changes whilst looking at this:

- I believe use of `pkg_resources` for loading images is deprecated, so updated that
- Pulled out graph layout functionality into a shared module
- Added a script (based on your notebook) to generate animations for all the environments to make testing easier

Wanted to check:

- Is it worth pulling some common functionality into the `Viewer` interface? There's common functionality for stuff like closing figures that is repeated in the implementations
- Do you feel all the viewers need updating in all cases? There's definitely better improved implementations for the viewers using the old functionality, but could pickup in further PRs if they are currently working otherwise.

sash-a:
Hi @zombie-einstein sorry for not having a look at this, the last two weeks have been a bit crazy

> Is it worth pulling some common functionality into the Viewer interface? There's common functionality for stuff like closing figures that is repeated in the implementations

I think that would be nice, but also happy to do this in a subsequent PR?

> Do you feel all the viewers need updating in all cases? There's definitely better improved implementations for the viewers using the old functionality, but could pickup in further PRs if they are currently working otherwise.

The main reason I'd like to do this is so that we don't pin to an old matplotlib version anymore so I do think we should update it in all cases, but if it's a bit slower than the old functionality I wouldn't worry because the viewers don't need to be too fast

examples/animation_test.py
sash-a, line 1: Since this is in examples can you rather call this `visualize_random_agent.py`

jumanji/environments/routing/cvrp/viewer.py
```diff
@@ -31,8 +35,9 @@ class CVRPViewer(Viewer):
     FIGURE_SIZE = (10.0, 10.0)
     NODE_COLOUR = "black"
     COLORMAP_NAME = "hsv"
-    NODE_SIZE = 150
-    DEPOT_SIZE = 250
+    NODE_SIZE = 0.01
+    ROUTE_NODE_SIZE = 100
+    DEPOT_SIZE = 0.04
```
sash-a, line 40: Why did this need to change?

sash-a (commented):
This looks good to me. Just a few very minor changes. Can you please explain why there are so many changes for the graph envs? I get there needs to be changes on reset, but not sure why things changed outside of the `if not jnp.array_equal(...)` block? (Not saying that it's wrong just curious why)

jumanji/environments/routing/cvrp/viewer.py
zombie-einstein: So (somewhat annoyingly) if you draw nodes/points with a circle-artist the scale (if I remember rightly) is a fraction of the region (hence the fractional values), but using a scatter plot the larger integer value gives sensible node sizes. 
This could change though, I think it's generally sensible to draw all the points in one go using a scatter, rather than individual artists.

jumanji/environments/routing/cvrp/viewer.py
```diff
@@ -99,17 +105,45 @@ def animate(
...
+                for i in range(1, x_coords.shape[0]):
+                    nodes[i].set_center((x_coords[i], y_coords[i]))
+                updated.extend(nodes)
+
+            while routes:
+                quiver, route_nodes = routes.pop()
+                quiver.remove()
+                route_nodes.remove()
+                updated.append(quiver)
+                updated.append(route_nodes)
```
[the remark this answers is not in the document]
zombie-einstein: Yeah will do, I may add a longer comment on the interface outlining how some of the  animation functionality is implemented.

In this case the members of the routes, and number of nodes are not static, and so the artists representing them need to be fully replaced, by removing the current ones. You can't pass state (I think) as part of the argument to the update function and so need to keep references to the existing artists in the scope outside the update function. So in the above the steps are:

- Pop the current artists from the outer list
- Call the remove method to remove them from the plot
- Add them to the list of updated artists returned from the update function
- Create new artists and push them in the outer list for use in later updates

I've used a list to act as mutable storage for artists mainly for convenience, maybe a global or class variable would look nicer. Also quite a few lines of code here, was looking to review if there is a neater way to implement.

zombie-einstein:
> This looks good to me. Just a few very minor changes. Can you please explain why there are so many changes for the graph envs? I get there needs to be changes on reset, but not sure why things changed outside of the `if not jnp.array_equal(...)` block? (Not saying that it's wrong just curious why)

Sure, so a couple things for graphs in general:

 - For the graphs the reset generally redraws the full plot, and in some cases relatively heavy computation to layout the nodes (which is _REALLY_ slow if done every step). So the full redraw is only done on a new episode, and means removing all the existing artists and replacing them with the updated ones using some mutable storage for artists outside the animation update function.
 - In some cases once the nodes and edges are drawn you can just update their properties inside an episode, e.g. updating their colour without needing to redraw them.
 - In the worse cases the plot has features that vary in size every step, e.g. in crvp the route changes every step, so need to update all the artists, and so for convenience I've just been using the existing functionality to remove the existing artists and replace them with newly created ones.

Compared to other environments the graphs are a pain due to all the nodes/edges moving episode, and the potential variable number of artists each step. In comparison for something like Suduko the grid is static, and you know there is a fixed grid of numbers that can just be updated in the animation, you don't need to mess about adding and removing artists inside the update function!

zombie-einstein:
> Hi @zombie-einstein sorry for not having a look at this, the last two weeks have been a bit crazy
> 
> > Is it worth pulling some common functionality into the Viewer interface? There's common functionality for stuff like closing figures that is repeated in the implementations
> 
> I think that would be nice, but also happy to do this in a subsequent PR?

Cool, I might just grab it, as where implementations vary it kind of makes it hared to refactor when trying to work out what different implementations are doing.

> > Do you feel all the viewers need updating in all cases? There's definitely better improved implementations for the viewers using the old functionality, but could pickup in further PRs if they are currently working otherwise.
> 
> The main reason I'd like to do this is so that we don't pin to an old matplotlib version anymore so I do think we should update it in all cases, but if it's a bit slower than the old functionality I wouldn't worry because the viewers don't need to be too fast

Everything is functional now with the updated version (the graphs were really the ones that didn't play nicely with the new API) updates would be mostly about efficient use of new API, e.g. updating artists rather then continually redrawing. My main concern was a the size of the PR, but I can chip away at the other envs if you don't mind it.

sash-a:
> Compared to other environments the graphs are a pain due to all the nodes/edges moving episode, and the potential variable number of artists each step. In comparison for something like Suduko the grid is static, and you know there is a fixed grid of numbers that can just be updated in the animation, you don't need to mess about adding and removing artists inside the update function!

I see, thanks for the explainer :+1: 

> Everything is functional now with the updated version (the graphs were really the ones that didn't play nicely with the new API) updates would be mostly about efficient use of new API, e.g. updating artists rather then continually redrawing. My main concern was a the size of the PR, but I can chip away at the other envs if you don't mind it.

Honestly I'm quite happy with the PR as is and would prefer subsequent changes in a future PR? Could you make an issue with a brief explainer of the speed improvements and another for pulling out common functionality?

If you're happy with this I'm happy to merge the PR as is :smile:

zombie-einstein:
Thanks @sash-a sounds like a plan. I'd made a start on the pulling out common functionality so I'll roll that into this.

jumanji/environments/logic/graph_coloring/viewer.py
```diff
@@ -21,37 +21,37 @@
...
+        super().__init__(name, render_mode)
+
+    def render(self, state: State) -> Optional[NDArray]:
         self._clear_display()
         self._set_params(state)
         fig, ax = self._get_fig_ax()
         ax.clear()
         self._prepare_figure(ax, state)
-        if save_path:
-            fig.savefig(save_path, bbox_inches="tight", pad_inches=0.2)
```
sash-a, line 58: How does the saving work now?

jumanji/environments/routing/mmst/viewer.py
```diff
@@ -70,31 +65,24 @@ def __init__(
...
         Return:
             pixel RGB array
         """
-        num_nodes = state.adj_matrix.shape[0]
-        node_scale = 5 + int(np.sqrt(num_nodes))
-
         self._clear_display()
-        fig, ax = self._get_fig_ax(node_scale)
-        fig.subplots_adjust(left=0.1, right=0.9, bottom=0.1, top=0.9)
+        # fig, ax = self._get_fig_ax(node_scale)
```
sash-a, line 80:
```suggestion
```

jumanji/environments/routing/mmst/viewer.py
```diff
@@ -239,10 +227,10 @@ def animate(
             Animation that can be saved as a GIF, MP4, or rendered with HTML.
         """
 
-        num_nodes = states[0].adj_matrix.shape[0]
-        node_scale = 5 + int(np.sqrt(num_nodes))
-        fig, ax = plt.subplots(num=f"{self._name}Animation", figsize=(node_scale, node_scale))
-        plt.close(fig)
+        # num_nodes = states[0].adj_matrix.shape[0]
+        # node_scale = 5 + int(np.sqrt(num_nodes))
```
sash-a, line 231: I assume this was just testing?

```suggestion
```

sash-a (commented):
Thanks for all the effort, really nice reduction in complexity for most of the viewers! Just have a couple questions and then I need to run it on my side to check everything still renders correctly :smile:

zombie-einstein:
> Thanks for all the effort, really nice reduction in complexity for most of the viewers! Just have a couple questions and then I need to run it on my side to check everything still renders correctly 😄

Thanks @sash-a, I do need to a final pass over this PR. A few of the layouts of the plots need updating now the axes creation has been consolidated, and remove the commented blocks I was just holding onto whilst I complete the tweaks.

zombie-einstein:
Hi @sash-a , I think this is sorted now if you wanted to grab a look. It'd be great if you could see the animation and renders make sense.

I try do a final check that things still work Ok in notebooks.

jumanji/environments/__init__.py
```diff
@@ -82,7 +82,7 @@ def is_notebook() -> bool:
         if is_colab():
             backend = "inline"
         elif is_notebook():
-            backend = "notebook"
+            backend = "ipympl"
```
zombie-einstein, line 85: I think this is not the correct way to update the render in a jupyter notebook (as mentioned in #272). I've tested this locally with what I guess is fairly recent version of jupyter. Only potential issue is if this breaks behaviour of older versions of notebooks?

Without this change I just get the message  `Javascript Error: IPython is not defined` when trying to call the render method in a notebook.

jumanji/viewer.py
```diff
@@ -60,3 +64,106 @@ def close(self) -> None:
...
+                - "human": render the environment on screen.
+                - "rgb_array": return a numpy array frame representing the environment.
+            figure_size: Figure size, default (10.0, 10.0)
+        """
+        self._name = name
+        self._animation: Optional[animation.Animation] = None
+        self.figure_size = figure_size
+
+        # Render interactive animations in Jupyter
+        plt.rcParams["animation.html"] = "jshtml"
```
zombie-einstein, line 95: Setting this displays animations nicely as interactive inline animations in Jupyter notebooks. I'm not totally sure this has no side-effects for in other cases (saving animations as gifs seems fine) so it seemed safest to stick it here  where we know the user is using this style of matplotlib viewer.

The other option could be just to leave the rendering of animations in notebooks to the user, but seems more annoying to have to find this information out than just setting it somewhere in the package.

sash-a: Looking around the internet I think this is fine. If people find a side effect we can modify it

zombie-einstein:
> Ok I can confirm that all animations work for me and when testing in a notebook it also seems to work fine, I just needed to install `ipympl` which I think is expected if you want to use jupyter.
> 
> So are you happy to merge otherwise?

Great, glad it was not only machine that worked! 

Is it easy for you to try it out on Colab? That's the only thing I didn't get round to testing out. But otherwise sure think this is goo to merge if you are happy.

Once it's merged I'll make that issue with some details of upgrading remaining viewers.

zombie-einstein:
Oh and just remembered to check if you had any thoughts on passing a save path to the render method on the viewer? They should all be consistent now (with no save path), but not sure if the intention was to allow saving across all the viewers.

sash-a:
Trying this on colab the rendering seems to work perfectly, but I have an issue with evosax (dependency of esquilax), they made a new release yesterday so it's likely a bug from that. I manually downgraded to `evosax<0.2.0` and it seems to have worked fine. Not quite sure what the issue is though

zombie-einstein:
Oh yeah looking at there's been some change to the API between versions, so will constrain it to the working version.

sash-a (dismissed):
Just need you to fix the failing CI, but approving in the meantime :fire:

sash-a:
> Oh I'll try grab a look. I'm was looking to make this and a couple dependencies optional, as they are not totally necessary to use the package.

I think optional would be ideal for jumanji and our wider RL/MARL ecosystem if possible. But if this is too much of a mission just pinning it would be ok too :smile: 

> Are you happy to merge this (once the interface is updated)? I could look to pin Evosax requirement to the working version in the meantime if it's a persistent issue.

Yes!

zombie-einstein:
> Just need you to fix the failing CI, but approving in the meantime 🔥

Can you re-run it? I think this is a combination of some network failure with hugging-face, and also the issue I've raised #277 with latest Jax version

> I think optional would be ideal for jumanji and our wider RL/MARL ecosystem if possible. But if this is too much of a mission just pinning it would be ok too 😄

Made a start on this, a bit of work to figure out the details of the optional features, but should have out soonish.

sash-a:
Should be fixed in #278. 

Basically some of our tests relied on the state of the random generator which broke when jax changed the output of random calls in 0.5. Should be merged tomorrow just waiting for a review from another maintainer :smile:

jumanji/environments/packing/job_shop/viewer.py
```diff
@@ -45,24 +45,25 @@ def __init__(
...
         self._max_op_duration = max_op_duration
 
         # Have additional color to avoid two jobs having same color when using hsv colormap
-        self._cmap = matplotlib.cm.get_cmap(self.COLORMAP_NAME, self._num_jobs + 1)
+        self._cmap = plt.get_cmap(self.COLORMAP_NAME, self._num_jobs + 1)
 
-        # The animation must be stored in a variable that lives as long as the
-        # animation should run. Otherwise, the animation will get garbage-collected.
-        self._animation: Optional[matplotlib.animation.Animation] = None
+        super().__init__(name, "human")
```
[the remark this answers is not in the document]
sash-a: @zombie-einstein thought I'd try this out, seems like jobshop doesn't have a render mode arg? Other than this I think we're ready to merge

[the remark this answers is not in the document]
zombie-einstein: Oh yeah, I think originally the render method was hard-coded to human rendering, so to play it safe I think I hard-coded it here. It can't hurt to pass the value in though. I'll pick it up today.

sash-a (approved):
:1st_place_medal:

Issue #277, opened 2025-03-12
bug: Jax>0.5 issues?


zombie-einstein:
### Description

You see any issues w `jax>0.5?` I updated my local installation and now see a bunch of test failures, e.g.:

```
FAILED jumanji/environments/logic/minesweeper/env_test.py::test_minesweeper__done_invalid_action - assert 1 == 2
```

Manually reinstalling a version below `0.5` using 

```
pip install -U "jax[cuda12]==0.4.36"
```

seems to work fine?

### What Jumanji version are you using?

Main branch

### Which accelerator(s) are you using?

GPU

zombie-einstein:
I'd guess Esquilax was pinning the version below 0.5, and now that has been relaxed this issue has popped up.

PR #278, opened 2025-03-17
chore: fixes to tests for JAX 0.5 upgrade

sash-a:
Jax 0.5 changed the output of random calls, so the same seed would output different numbers. Thus I needed to change some of the tests as they relied on the state of the random generation of the environment. 

Some of the envs had bad tests that tested the exact values of the state/obs I have changed these to check for shape/dtype/min/max instead

Closes: #277


jumanji/environments/logic/minesweeper/env_test.py
```diff
@@ -38,7 +38,7 @@ def play_and_get_episode_stats(
     time_limit: int,
     force_start_state: Optional[State] = None,
 ) -> Tuple[List[float], List[StepType], int]:
-    state, timestep = jax.jit(env.reset)(jax.random.PRNGKey(0))
+    state, timestep = jax.jit(env.reset)(jax.random.PRNGKey(1))
```
WiemKhlifi, line 41: Is there a reason for changing the seed? Or is it related to some test values?

jumanji/environments/routing/multi_cvrp/env_test.py
```diff
@@ -230,8 +226,11 @@ def test_multicvrp__state_to_observation_timestep(self, multicvrp_env: MultiCVRP
         )
 
         # Check that the node demands and action masks are correct.
-        assert np.array_equal(obs.nodes.demands, test_node_demand)
-        assert np.array_equal(obs.action_mask, test_action_mask)
+        assert obs.nodes.demands.shape == (multicvrp_env._num_customers + 1,)
+        assert obs.action_mask.shape == (
+            multicvrp_env._num_vehicles,
```
WiemKhlifi, line 231: 🧠 💯

jumanji/environments/routing/maze/env_test.py
```diff
@@ -148,23 +148,29 @@ def test_maze__step(self, maze: Maze) -> None:
         assert state.agent_position == Position(row=2, col=2)
 
     def test_maze__action_mask(self, maze: Maze) -> None:
-        key = jax.random.PRNGKey(0)
-        state, _ = maze.reset(key)
-
+        walls = jnp.array(
```
WiemKhlifi, line 151: Is the reason for this to increase speed? I know the goal is to check action mask so initing the env is useless 🤔

WiemKhlifi (commented):
Thanks @sash-a for fixing it! LGTM just have some comments 🙌

jumanji/environments/routing/maze/env_test.py
sash-a: Similar to the other one, because 0.5 changes the output of seeds the reset produces a different state. These walls are just copying the reset walls from 0.4 JAX

WiemKhlifi (approved):
LGTM thanks @sash-a for these fixes 🥳

PR #280, opened 2025-05-06
build: add .gitmessage and conventional-pre-commit in pre-commit

MarcBoelle3:


MarcBoelle3:
Add .gitmessage and conventionl pre-commit in .pre-commit-config.yaml

Issue #101, opened 2023-03-28
docs: show test coverage in badge


dluo96:
### Is your feature request related to a problem? Please describe
We should display the test coverage in a badge on the README. This is important to show since **well-tested API** is a goal.

PR #287, opened 2025-10-11
docs: add test coverage badge (#101)

ayoub3bidi:
The implementation leverages existing infrastructure and adds minimal complexity while providing valuable visibility into test coverage. The badge will automatically update with each CI run, providing real-time coverage information to contributors and users.


sash-a:
Thanks, but I think we will skip adding this badge, tests passing is enough for us

PR #299, opened 2026-06-02
fix: correct player position spec bounds

nvdaz:
correct the x/y in player position spec in the pac man environment, which is inexplicably reversed currently


sash-a (approved):
Thanks for spotting this bug!

PR #301, opened 2026-06-07
fix: pac man custom time limits

nvdaz:
fix the custom time limit option for pac man, which is currently always set to 1000 regardless of how it is configured


sash-a:
Ah this is quite a subtle bug, but something we do throughout the codebase. The `or` should check the `None` and return the default value if `time_limit` is `None`, but they are in the wrong order, it should be `time_limit or 1000`.

Thanks for spotting it!

I want to merge in #300 first as that will fix our CI, just need some time to test it :smile:

nvdaz:
> Ah this is quite a subtle bug, but something we do throughout the codebase. The `or` should check the `None` and return the default value if `time_limit` is `None`, but they are in the wrong order, it should be `time_limit or 1000`.

Is the suggestion here to use `time_limit or 1000` instead of `time_limit=1000`? I usde `=1000` since it seems more readable and (looking quickly) seems to be more common in [other envs](https://github.com/instadeepai/jumanji/blob/a38920a49171011ab6640a11cae69a496d0c3a9a/jumanji/environments/routing/robot_warehouse/env.py#L140) 


> I want to merge in #300 first as that will fix our CI, just need some time to test it 😄

sure, this is great!

sash-a:
No, no suggestion, I prefer your change, much easier to read

PR #303, opened 2026-06-23
Use functools.partial instead of lambdas for lax.cond branches (#245)

### LINKED ISSUE #245: Proposal for Code Structure Improvement Using `jax.lax.cond`

This is a simple question regarding code style. It is not related to any bugs.

https://github.com/instadeepai/jumanji/blob/fd511b4523425bdcc69f2fd3c8869cefddd49f1d/jumanji/environments/routing/connector/env.py#L184-L198

https://github.com/instadeepai/jumanji/blob/fd511b4523425bdcc69f2fd3c8869cefddd49f1d/jumanji/environments/logic/game_2048/env.py#L222-L236

Rather than repeatedly using `lambda` and duplicating variables as shown in the code above, it seems better to follow the functional style of `jax.lax.cond` and write it in the style of the solution code below.

It seems that there is little to no difference in performance.
If this is a minor issue, I will close it.

```python
timestep = jax.lax.cond(
    done,
    termination,
    transition,
    reward,
    observation,
    extras,
)
```

### Describe alternatives you've considered
None

### Additional context

https://github.com/instadeepai/jumanji/blob/fd511b4523425bdcc69f2fd3c8869cefddd49f1d/jumanji/environments/logic/minesweeper/env.py#L178-L185

https://github.com/instadeepai/jumanji/blob/fd511b4523425bdcc69f2fd3c8869cefddd49f1d/jumanji/environments/logic/graph_coloring/env.py#L202-L209

https://github.com/instadeepai/jumanji/blob/fd511b4523425bdcc69f2fd3c8869cefddd49f1d/jumanji/environments/logic/rubiks_cube/env.py#L169-L176

https://github.com/instadeepai/jumanji/blob/fd511b4523425bdcc69f2fd3c8869cefddd49f1d/jumanji/environments/logic/sudoku/env.py#L124-L132

https://github.com/instadeepai/jumanji/blob/fd511b4523425bdcc69f2fd3c8869cefddd49f1d/jumanji/environments/packing/tetris/env.py#L235-L242

---

### Misc
- [X] Check for duplicate requests.

Elgoghel:
Addresses #245. Replaces the lambda-wrapped jax.lax.cond branches with the
functools.partial direct-call style already used in search_and_rescue, in four
environments: game_2048, sliding_tile_puzzle, bin_pack, and sokoban. reward and
observation are passed as cond operands and extras is bound via partial.

Behaviour preserving: termination has no `discount` parameter while
transition/truncation do, so extras is bound by keyword to keep both branches
correct. All existing tests for the four environments pass locally (35 passed).

Note: connector is intentionally left unchanged. Its lax.switch term_fn adapts
termination's differing signature (it drops discount), so that lambda is
load-bearing rather than redundant. Happy to revisit if a named helper is preferred.

### DIFF
--- a/jumanji/environments/logic/game_2048/env.py
+++ b/jumanji/environments/logic/game_2048/env.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from functools import cached_property
+from functools import cached_property, partial
 from typing import Optional, Sequence, Tuple
 
 import chex
@@ -217,16 +217,10 @@ class Game2048(Environment[State, specs.DiscreteArray, Observation]):
         extras = {"highest_tile": highest_tile}
         timestep = jax.lax.cond(
             done,
-            lambda: termination(
-                reward=reward,
-                observation=observation,
-                extras=extras,
-            ),
-            lambda: transition(
-                reward=reward,
-                observation=observation,
-                extras=extras,
-            ),
+            partial(termination, extras=extras),
+            partial(transition, extras=extras),
+            reward,
+            observation,
         )
 
         return state, timestep
--- a/jumanji/environments/logic/sliding_tile_puzzle/env.py
+++ b/jumanji/environments/logic/sliding_tile_puzzle/env.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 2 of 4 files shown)


sash-a (approved):
Thanks for this, much cleaner 😄

PR #310, opened 2026-08-24
feat: expose observation from state in public API

quinnarnold:
## Summary

Add a public `Environment.observe(state)` method for deriving an observation from an environment state.

- implement the method for all built-in environments by promoting existing private helpers or extracting existing inline observation construction
- delegate through `Wrapper` and vectorize batched states in `VmapWrapper`
- verify reset and step observations against `observe(state)` through the shared environment checks
- expose the method in the environment API documentation

`Environment.observe` is abstract, consistent with `reset` and `step`. The shared environment check uses a JIT-compiled observe function for both reset and step observations.

Merged current `main`, preserving the Tetris observation counter fix from #313.

Closes #195.

## Testing

Latest revision:
- Focused Tetris, shared smoke-helper, fake-environment, and wrapper tests: `69 passed`.
- Environment smoke selection: `55 passed`.
- All pre-commit checks passed.

Original implementation validation:

- `uv run pre-commit run --all-files`
- environment smoke selection with the new shared assertions (`55 passed`)
- `uv run pytest -q jumanji/wrappers_test.py` (`52 passed`)
- full suite (`816 passed`, `3 skipped`; one unrelated existing Search-and-Rescue floating-point boundary test failed)
- documentation build

AI-assisted: Yes.

PR #313, opened 2026-09-03
fix(tetris): observation step_count is always 0

yurekami:
`Tetris.step` builds the next observation with a hard-coded `step_count=jnp.array(0, jnp.int32)` even though `next_state.step_count` has just been incremented, so `Observation.step_count` (documented as the "current number of steps in the episode") is 0 at every step. The Tetris actor-critic network feeds `observation.step_count / time_limit` into its torso, so that input is currently a constant.

This uses the incremented `step_count` in the observation. Since the episode ends when `step_count >= time_limit`, the terminal observation carries `step_count == time_limit`, so the spec is widened to `DiscreteArray(time_limit + 1)` (matching the inclusive `maximum=self.time_limit` bound the other time-limited envs use) so that the terminal observation validates.

Adds `test_tetris__observation_step_count`, which walks valid actions to the time limit, checks the observation counter matches the state's, and validates every observation against the spec (fails on `main` at the first step with `0 == 1`).

Note: #310 rewrites the same `next_observation` block into `Tetris.observe`, and its version still hard-codes `step_count=0`. If #310 lands first, this becomes a one-line change there plus the same spec bump; happy to rebase either way. `Snake` has the same spec off-by-one (`DiscreteArray(self.time_limit)` while the terminal observation carries `time_limit`), which I can send as a follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_011vZKTbRaDh8FuXSbVpQomD


sash-a (approved):
Thanks for the bugfix

PR #316, opened 2026-09-09
feat(registration): default to latest environment version

### LINKED ISSUE #94: feat(registration): default to latest version
`jumanji.make("TSP")` should default to latest version of TSP, e.g. "TSP-v1". This way, a user who sees the codebase and agrees with the current version of TSP should be able to use TSP and not have to search for what version of TSP we are at.

yurekami:
`jumanji.make("TSP")` currently raises a missing-version error. Allow bare environment names to select the highest registered numeric version, as requested in #94.

Resolution is limited to `make()`: registration still requires versioned IDs, and explicit requests keep selecting exactly the requested version. Constructor arguments continue through the existing override path. The registration guide includes the new usage and explains how to pin a version.

Fixes #94.

Validation on Windows with Python 3.13 and the locked dependencies:

- Before the change: all three new latest-version regression cases failed with the missing-version error.
- `python -m pytest jumanji/registration_test.py -k 'not test_registration__make' -q`: 15 passed, 1 deselected. The existing all-environments test was excluded because it downloads the Sokoban dataset.
- Real `make`/`reset` checks passed for `BinPack`, `BinPack-v2`, `TSP`, and `TSP-v1` using `jax.random.PRNGKey(0)`.
- All applicable pre-commit checks passed for the three changed files, including Ruff, the full-project ty check, whitespace/line endings, and license checks.


jumanji/registration.py
```diff
@@ -151,7 +151,18 @@ def load(name: str) -> Type[Environment]:
...
+    """Create an environment, defaulting to its latest registered version."""
+    env_id = id
+    match = ENV_NAME_RE.fullmatch(id)
+    if match and match.group("version") is None:
+        latest_version = max(
+            (spec.version for spec in _REGISTRY.values() if spec.name == id),
+            default=None,
+        )
+        if latest_version is not None:
+            env_id = get_env_id(id, latest_version)
```
sash-a, line 163: If there is no `latest_version` we should raise an error, please also test this case

jumanji/registration.py
```diff
@@ -151,7 +151,18 @@ def load(name: str) -> Type[Environment]:
 
 
 def make(id: str, *args: Any, **kwargs: Any) -> Environment:
-    env_name, version = parse_env_id(id)
+    """Create an environment, defaulting to its latest registered version."""
+    env_id = id
```
sash-a, line 155: Why create this alias, I think we can just use `id` throughout

sash-a (commented):
Thanks for this, just a few small changes

jumanji/registration.py
yurekami: Done in 24346f2. `make` now raises `ValueError("Unregistered environment Fake: no version of it is registered. Please select from the registered environments: ...")` when the unversioned name has nothing registered, instead of falling through to `parse_env_id`'s "Version missing" message. `test_make__no_registered_version` registers only `Fake-other-v0` and checks that `make("Fake")` raises with that message; the `-other` name is there so a prefix match does not count as a version.

jumanji/registration.py
yurekami: Removed in 24346f2. The versionless branch now sets `env_name, version` directly and the versioned branch goes through `parse_env_id(id)`, so there is no alias; `env_id` is only the resolved `get_env_id(env_name, version)` as before. I did not reassign `id` itself because ruff flags that as A001.

